PR #1335, opened 2014-12-23
Allow OrderPayment transition from settled to charged_back

rollick:


rollick:
Merged this into saas branch.

PR #1339, opened 2014-12-24
added logic language + style

JhonM:


aethemba:
@JhonM  , is this supposed to go into develop, or should it go into saas?

JhonM:
In develop initially but we can moved it to Saas.

aethemba:
@JhonM , was this specifically done for Booking? If so, we should move it into develop, otherwise Saas would be the appropriate branch.

JhonM:
Booking don't have multiple languages right? So it can be moved to Saas.

aethemba:
yeah

PR #1378, opened 2015-01-19
Feature/bb 2377 sign up in

gannetson:


rollick:
@gannetson Is this working as you would expect? What happens when the confirmation email is different to the email?

PR #1395, opened 2015-01-29
Feature/bb 2375 payment settings

gannetson:


bluebottle/payments/services.py
```diff
@@ -1,6 +1,6 @@
 from bluebottle.payments.exception import PaymentException
 import re
-from django.conf import settings
+from bluebottle.clients import properties
```
rollick: I thought the idea was to be able to extend the django settings based on the client. Shouldn't be able to do getattr(settings, 'PAYMENT_METHODS', ()) so that applications which use settings will "just work".

gannetson: What I got from Ivo is that injecting properties into your settings isn't good. 
@iivvoo could you shed some light on this?

iivvoo: settings aren't thread safe so modifying it one one request may affect a parallel request. And modifying settings in general is tricky anyway - you need to be careful to restore previous values, (e.g. exceptions) etc.

properties will default back to settings if a value is not found in the client's properties. But you *will* need to access them through bluebottle.clients.settings - there's no (sane) way to support settings in a multi tenant setup

gannetson: @iivvoo do you mean bluebottle.clients.properties instead of bluebottle.clients.settings?

iivvoo: Yeah :) (fixed)

PR #1403, opened 2015-02-10
Feature/bb 2492 mails tenant specific

iivvoo:


bluebottle/clients/context.py
```diff
@@ -0,0 +1,12 @@
...
+
+class ClientContext(Context):
+    """
+    A Context with builtin properties support. Mostly meant
+    for rendering mail templates that don't get the
+    ContextProcessor goodness
+    """
+    def __init__(self, *args, **kw):
+        super(ClientContext, self).__init__(*args, **kw)
+        self.update({'tenant_properties':properties})
```
rollick: Should we pass the whole context to the template? This also includes secret keys etc.

iivvoo: Good point. I'll restrict it to TENANT_MAIL_PROPERTIES for now.

PR #1421, opened 2015-02-18
Change how themes/phases are loaded

gannetson:


rollick:
We should do something like this for the default pages, skills and statistics data.

PR #1426, opened 2015-02-19
Clean up initial data and fix project realised phase BB-2550

rollick:


bluebottle/test/utils.py
```diff
@@ -49,9 +49,12 @@ def css_dict(style):
 class InitProjectDataMixin(object):
 
     def init_projects(self):
+        from django.core import management
         """
         Set up some basic models needed for project creation.
         """
+        management.call_command('loaddata', 'project_data.json', verbosity=0)
```
gannetson: I think this should already be loaded by migration
https://github.com/onepercentclub/bluebottle/blob/saas/bluebottle/bb_projects/migrations/0003_project_phases_and_themes.py#L12

rollick: I think it is faking those migrations then because without the mangement command the project phases don't get loaded in the tests.

PR #1445, opened 2015-02-26
check if there is a default signup language key in client properties and...

aethemba:
... use that value when creating a user

### DIFF
--- a/bluebottle/bb_accounts/views.py
+++ b/bluebottle/bb_accounts/views.py
@@ -60,10 +60,14 @@ class UserCreate(generics.CreateAPIView):
         return "Users"
 
     def pre_save(self, obj):
-        if hasattr(properties, 'DEFAULT_SIGNUP_LANGUAGE'):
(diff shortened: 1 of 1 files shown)


bluebottle/bb_accounts/views.py
```diff
@@ -59,7 +60,10 @@ def get_name(self):
         return "Users"
 
     def pre_save(self, obj):
-        obj.primary_language = self.request.LANGUAGE_CODE[:2]
+        if hasattr(properties, 'DEFAULT_SIGNUP_LANGUAGE'):
```
rollick: I don't think this is correct. We should check if obj.primary_language is in the list of supported LANGUAGES: if it is then set it for the user, otherwise use the LANGUAGE_CODE. 

LANGUAGES and LANGUAGE_CODE are in the tenant properties.

aethemba: Yeah makes sense, good point. Ill update the PR this weekend or maybe tonight.

PR #1469, opened 2015-03-11
Make sure we can switch to live payments

gannetson:


bluebottle/payments_docdata/adapters.py
```diff
@@ -54,6 +54,7 @@ class DocdataPaymentAdapter(BasePaymentAdapter):
     }
 
     def __init__(self, *args, **kwargs):
+        self.live_mode = properties.LIVE_PAYMENTS_ENABLED
```
rollick: Should this be a getattr with a False default?

bluebottle/payments_docdata/adapters.py
```diff
@@ -221,21 +220,18 @@ def create_payment(self):
...
+        # or fallback on the default LANGUAGE_CODE in settings
         
         user_language = self.order_payment.order.user.primary_language
         
         if user_language:
             client_language = user_language
         else:
-            # Language code is a default Django setting. We assume the key exists.
-            client_language = properties.LANGUAGE_CODE
+            client_language = settings.LANGUAGE_CODE
```
rollick: properties..

PR #1478, opened 2015-03-16
make two addtional strings translatable

aethemba:


bluebottle/bb_donations/donationmail.py
```diff
@@ -15,13 +15,22 @@ def successful_donation_fundraiser_mail(instance):
 
     fundraiser_link = '/go/fundraisers/{0}'.format(instance.fundraiser.id)
 
+    if instance.fundraiser.owner.email:
```
gannetson: Won't this throw an error if there's no fundraiser?

aethemba: There's a check for that a few lines above

PR #1487, opened 2015-03-18
Handle SMTP errors gracefully

iivvoo:


bluebottle/mail/__init__.py
```diff
@@ -27,4 +32,7 @@ def send_mail(template_name, subject, to, **kwargs):
     msg = EmailMultiAlternatives(subject=subject, body=text_content, to=[to.email])
     msg.attach_alternative(html_content, "text/html")
 
-    return msg.send()
+    try:
+        return msg.send()
+    except smtplib.SMTPException:
```
gannetson: What will happen if one of the log handlers tries to send a mail?

iivvoo: As long as it does not use bluebottle.mail for that, nothing will change.

PR #1488, opened 2015-03-18
handle wallposts with deleted projects/fundraisers etc

iivvoo:


bluebottle/wallposts/models.py
```diff
@@ -156,4 +156,3 @@ def __unicode__(self):
 #Import the signals for sending mails in case they are present
 if 'bluebottle.tasks' in settings.INSTALLED_APPS and 'bluebottle.projects' in settings.INSTALLED_APPS:
     import mails
-
```
rollick: @iivvoo Don't we need a blank line for pep8??

PR #1489, opened 2015-03-20
Feature/bb 2935

aethemba:


bluebottle/bb_accounts/utils.py
```diff
@@ -17,35 +15,29 @@ def send_welcome_mail(user=None):
     else:
         translation.activate(properties.LANGUAGE_CODE)
 
-    c = ClientContext({
+    c = {
```
gannetson: If we make changes here maybe use a longer variable name?

bluebottle/bb_accounts/views.py
```diff
@@ -174,7 +187,8 @@ def save(self, password_reset_form, domain_override=None,
                 domain = tenant_url()
             else:
                 site_name = domain = domain_override
-            c = ClientContext({
+
+            c = {
```
gannetson: See above. Longer var names are preferred

bluebottle/bb_projects/projectwallmails.py
```diff
@@ -17,13 +19,17 @@ def notify(self):
         project = self.post
         project_owner = project.owner
 
-        # Implement 1a: send email to Object owner, if Wallpost author is not the Object owner.
+        # Implement 1a: send email to Object owner, if Wallpost author is not
+        # the Object owner.
         if self.author != project_owner:
+            self.activate_language(project_owner)
+            subject = _('%(author)s commented on your project') % {
+                'author': self.author.get_short_name()}
```
gannetson: I think we should move to this format:

`subject = _('{author} commented on your project').format(author=self.author.get_short_name())`

That do you think? Not necessary for this PR though

PR #1513, opened 2015-04-07
Fix changing payment method.

gannetson:
This also fixes moving a order(payment) from failed to pending/success.
Also stores the payment transactions.

This needs thorough testing and probably some nice automated tests.


bluebottle/payments_logger/tests/test_unittests.py
```diff
@@ -48,7 +48,7 @@ def test_check_authorized_status_logged(self, mock_fetch_status, mock_transactio
 
         # Check that the status change was logged
         self.assertEqual(last_log.payment_id, self.order_payment.payment.id)
-        self.assertEqual(last_log.message, 'DocdataPayment object - a new payment status authorized')
+        self.assertEqual(last_log.message, 'DocdataPayment object - a new payment status started')
```
aethemba: I think you forgot to add the change that updates the message. The tests fail because "authorized" is return instead of started.

PR #1515, opened 2015-04-08
Feature/bb 3230 docdata latest payment

eodolphi:
To check the status of a payments, we should look at the totals, instead of looking at the status of the first payment.

This branch fixes that, reenables the storing of transactions, and adds a bunch of tests for payments status response edge cases.


aethemba:
This will pass when you merge the fixed PR of @gannetson  that includes a simple text change.

bluebottle/payments_docdata/adapters.py
```diff
@@ -285,13 +266,40 @@ def check_payment_status(self):
...
+                # Everything was captured
+                status = StatusDefinition.SETTLED
+            elif int(totals.totalAcquirerApproved) + int(totals.totalAcquirerPending) + int(totals.totalShopperPending) == int(totals.totalRegistered):
+                # Everything was authorized
+                status = StatusDefinition.AUTHORIZED
+            else:
+                # Anything else (Partly captured, partly charged back, etc)
+                status = StatusDefinition.UNKNOWN
+
+        if self.payment.status != status:
```
gannetson: Maybe we should always update this, even if the status doesn't change?
Are you sure this doesn't change while status doesn't?

eodolphi: I was wondering about that to. Kept it the same for now. If this is not a performance problem, then yes let's change this just to be sure the data is correct.

gannetson:
@eodolphi could you merge saas branch so this can be pulled cleanly?

eodolphi:
@gannetson Done!

PR #1538, opened 2015-04-20
Change member migration to depend on geo location

gannetson:


rollick:
@aksel has already updated this in a different PR.

PR #1543, opened 2015-04-21
Feature/bb cronsalesforce

aethemba:


gannetson:
To enable this we have to set `CELERYBEAT_SCHEDULE` in reef, correct?

PR #1555, opened 2015-04-29
Remove old project bank fields.

gannetson:


bluebottle/projects/tests/test_api.py
```diff
@@ -403,21 +379,21 @@ def test_set_bank_details(self):
...
+    #         'account_number': 'NL18ABNA0484fesewf869868',
+    #     }
+    #
+    #     response = self.client.post(self.manage_projects_url, project_data,
+    #                                 token=self.some_user_token)
+    #
+    #     self.assertEquals(response.status_code,
+    #                       status.HTTP_400_BAD_REQUEST)
+    #     self.assertEquals(json.loads(response.content)['account_number'][0],
+    #                       'Wrong IBAN length for country code NL.')
```
eodolphi: Shouldn't we keep this test? If not it should just be removed.

bluebottle/bb_projects/models.py
```diff
@@ -193,24 +193,11 @@ class BaseProject(models.Model, GetTweetMixin):
         related_name="project_account_holder_country")
 
     # Bank details
-    account_iban = IBANField(_("account IBAN"), null=True, blank=True)
+    account_number = models.CharField(_("Account number"), max_length=255, null=True, blank=True)
```
eodolphi: This also requires a change in reef right?

bluebottle/bb_projects/serializers.py
```diff
@@ -207,7 +201,5 @@ class Meta:
                   'organization', 'language', 'account_holder_name',
                   'account_holder_address', 'account_holder_postal_code',
                   'account_holder_city', 'account_holder_country',
-                  'account_iban', 'account_bic', 'account_number',
-                  'account_bank_name', 'account_bank_address',
-                  'account_bank_postal_code', 'account_bank_city',
-                  'account_bank_country', 'account_other', 'tasks')
+                  'account_number', 'account_bic',
+                  'account_bank_country', 'tasks')
```
eodolphi: Do we still need the country? Can that be different from account_holder_country?

bluebottle/bb_projects/models.py
gannetson: You already did the changes to reef, right?

bluebottle/bb_projects/serializers.py
gannetson: Yeah, country should be in as per Joris.

bluebottle/bb_projects/models.py
eodolphi: Yeah, nut it does not match exactly. There it is called account_iban. Which is probably wrong.

gannetson: Yes because it can be a non-iban too

PR #1566, opened 2015-05-04
Feature/bb 3444 revert website field

eodolphi:


eodolphi:
For people who already migrated their database to the current saas branch: You will need to dd the website column back to your database:

ALTER TABLE organizations_organization ADD COLUMN website varchar(200);

PR #1590, opened 2015-05-21
New tenant extras with en-trans fix

gannetson:


aethemba:
@gannetson , if you fix the tests then I can merge this and the PR in reef in

PR #1592, opened 2015-05-26
Add authorized as a possible status for new payments (it can happen t…

eodolphi:
…hat nothing is captured, approved nor pending but the status is AUHTORIZED. Explicitly check for failed statuses otherwise set the status to UNKOWN


gannetson:
I think this PR should point to release/1205-hamill (the version now on production)
We should also update finate state machines for `OrderPayment` and `Order` to allow to change from `failed` to `settled` so we can set the records straight.

PR #1623, opened 2015-06-09
Merge release into saas

rollick:


eodolphi:
Hard to tell if there is anything missing. But I do not see anything weird.

PR #1630, opened 2015-06-12
Set newsletter default to True

gannetson:


gannetson:
Is this merged/solved in another way @rollick

PR #1638, opened 2015-06-24
Remove tenant_static templatetag (it is also in django-tenant-extras)

eodolphi:


aethemba:
@eodolphi , i cant merge this

PR #1673, opened 2015-07-31
Disable sites framework

gannetson:


bluebottle/redirects/middleware.py
```diff
@@ -19,12 +19,12 @@ class RedirectFallbackMiddleware(object):
...
-            raise ImproperlyConfigured(
-                "You cannot use RedirectFallbackMiddleware when "
-                "django.contrib.sites is not installed."
-            )
+    # def __init__(self):
+    #     if 'django.contrib.sites' not in settings.INSTALLED_APPS:
+    #         raise ImproperlyConfigured(
+    #             "You cannot use RedirectFallbackMiddleware when "
+    #             "django.contrib.sites is not installed."
+    #         )
```
eodolphi: Why not completely remove this? Do we need it back quickly?

gannetson:
True @eodolphi. I forgot about it. I just commented it to see if tests would pass. It's gone now.

PR #1675, opened 2015-08-03
Make it possible to POST `null` as the address

eodolphi:


eodolphi:
It seem other fronted functionality does not depend on the address or at least not in way that breaks if there is not even an empty address.

PR #1676, opened 2015-08-03
Override view to reset admin password

gannetson:


aethemba:
@gannetson , is this ready to be merged? I remember you saying something at standup so I wasn't sure

gannetson:
Nope not yet

PR #1681, opened 2015-08-04
Possible fixes for incorrect payout rules BB-4342

rollick:


aethemba:
@rollick , some unit tests of the payouts are breaking

PR #1699, opened 2015-08-11
Feature/bb 4404

aethemba:


bluebottle/votes/serializers.py
```diff
@@ -0,0 +1,13 @@
+from bluebottle.votes.models import Vote
+from bluebottle.bb_accounts.serializers import UserPreviewSerializer
+from rest_framework import serializers
+
+
+class VoteSerializer(serializers.ModelSerializer):
+    id = serializers.CharField(read_only=True)
+    voter = UserPreviewSerializer(read_only=True)
+    project = serializers.CharField(read_only=True)
```
rollick: Should we use a project preview serializer here?

aethemba: I think that should be ok. @eodolphi , what do you think?

eodolphi: I purposely did it this way, since you have already retrieved the project. This saves quite a bit of space in the response.

PR #1717, opened 2015-08-18
Remove `adopted` from suggestion metrics

eodolphi:
The adopted field in the metrics was querying for projects with the state submitted and a project reference. All submitted projects have a project reference, so this status was redundant.


aethemba:
Im not entirely sure if this is correct. I think it's best to double check with Paul or Joris about this because "adopted" was a required state in Booking at the time. (This has probably changed but better safe than sorry I guess)

PR #1728, opened 2015-08-21
Pre - Django 1.8.4

gannetson:
This is just updating some packages so we could switch to recent Django version in the future.


jfterpstra:
Please note: the mentioned version 1.1.16 of django-bb-salesforce is no longer the most recent version

PR #1754, opened 2015-09-03
Let the frotnend determine the language. If not, fall back to the LAN…

aethemba:
…GAUGE code in the tenant properties. Only in user create

### DIFF
--- a/bluebottle/members/serializers.py
+++ b/bluebottle/members/serializers.py
@@ -178,6 +178,7 @@ class UserCreateSerializer(serializers.ModelSerializer):
     password = PasswordField(required=True, max_length=128)
     username = serializers.CharField(read_only=True)
     jwt_token = serializers.CharField(source='get_jwt_token', read_only=True)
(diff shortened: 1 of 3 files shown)


bluebottle/members/serializers.py
```diff
@@ -194,7 +195,7 @@ def validate_email_confirmation(self, attrs, source):
     class Meta:
         model = BB_USER_MODEL
         fields = ('id', 'username', 'first_name', 'last_name',
-                  'email', 'password', 'jwt_token')
+                  'email', 'password', 'jwt_token', 'primary_language')
```
gannetson: Do we still need this??

aethemba: Yeah, the frontend determines the primary_language depending on what language the using is currently using. It posts this to the backend. (And without it we default to the  LANGUAGE_CODE in properties.py)

gannetson: Ok. Awaiting tests now ;-)

PR #1758, opened 2015-09-04
Add a separate boolean to mark co-financers.

eodolphi:
Before we used the user type, but since that was used before already,
and it might return, we use a different flag.

This also gives a bit more flexibility. Not all companies might have
been fit to show as co-financers.


aethemba:
@eodolphi , can you fix the conflict, then Ill merge the PR

PR #1767, opened 2015-09-17
Feature/bb 4651 login using torii

gannetson:


aethemba:
@gannetson , volgens mij kan deze er wel in he?

gannetson:
This will break the current Facebook login. It shou;d only go in if we merge the torii stuff in Reef, I think.

PR #1769, opened 2015-09-22
Do not count tasks with deadlines in the past when counting registere…

eodolphi:
…d/required people


aethemba:
Good stuff. We'll get it right someday I guess hehe ;)

PR #1782, opened 2015-10-12
DON'T MERGE YET - Accounting (Maykin)

gannetson:


sergei-maertens:
I fixed the failing test, but now Travis/Github are throwing 429: Too many request errors, so the build is failing because of that.

PR #1794, opened 2015-10-30
Financial backoffice

gannetson:


bluebottle/accounting/admin.py
```diff
@@ -0,0 +1,456 @@
...
+                       'start_date', 'end_date', 'collected_amount',
+                       'payments_count', 'payments_total', 'fee_total', 'costs_total', 'vat_costs',
+                       'local_payments_total', 'local_payments_count']
+
+    fieldsets = (
+        (None, {
+            'fields': ('payout_reference', 'payout_date', 'payout_amount',
+                       'start_date', 'end_date', 'collected_amount')
+        }),
+        ('Calculated from remote payments', {
```
eodolphi: This should probably be wrapped in a gettext call.

bluebottle/accounting/admin_views.py
```diff
@@ -0,0 +1,374 @@
...
+            payment.save()
+
+            # update/create the required payout
+            ProjectPayout = get_project_payout_model()
+            project = donation.project
+            project.update_amounts()
+            payouts = ProjectPayout.objects.filter(project=project)
+
+            # check the payouts and only update 'new' payouts, else create a new payout
+            if not payouts.exists():
```
eodolphi: That logic is a little weird. Why not 
```
if payoutexists:
    updateable = ....
```

And then skip the else.

tests.txt
```diff
@@ -0,0 +1,62 @@
+======================================================================
```
rollick: Is this needed?

bluebottle/accounting/utils.py
```diff
@@ -0,0 +1,293 @@
...
+            return first.add(second)
+        else:
+            try:
+                return Decimal(str(first)) + Decimal(str(second))
+            except:
+                return None
+
+
+class mydict(dict):
+
```
eodolphi: This should have a docstring explaining what is different about this dict compared to a normal one. It also should have a name that explains that.

bluebottle/accounting/utils.py
```diff
@@ -0,0 +1,293 @@
...
+        if set(keys) != set(dict2.keys()):
+            return new_dict
+
+        for key in keys:
+            new_dict[key] = add_up(self[key], dict2[key])
+
+        return new_dict
+
+
+class mylist(list):
```
eodolphi: See above. Same thing holds here.

bluebottle/accounting/views.py
```diff
@@ -0,0 +1,241 @@
...
+
+        context.update({
+             'app_label': 'accounting',
+             'title': _('Accountancy Overview') + header,
+             'statistics': statistics,
+        })
+        return context
+
+
+class MultiTenantAccountingDashboardView(InitialDatesMixin, OnepercentOnlyPermissionMixin,FormView):
```
eodolphi: Misses a space.

bluebottle/accounting/views.py
```diff
@@ -0,0 +1,241 @@
...
+    """
+    Same as AccountingDashboardView, but for multiple tenants.
+    Could be subclassed, but the distinction should be very clear,
+    this view can contain data from all tenants, the other view can NOT.
+    """
+    template_name = 'admin/accounting/dashboard.html'
+    #template_name = 'admin/accounting/multiadmin/multi_tenant_dashboard.html'
```
eodolphi: This should probably be removed.

bluebottle/accounting/views.py
```diff
@@ -0,0 +1,241 @@
...
+
+                header = ' - {}'.format(tenant.name)
+            else:
+                header = ' - All tenants'
+                statistics = mydict()
+                totals = mydict()
+                all_tenants = {}  # contains data for each tenant when all are selected
+
+                for tenant in get_tenant_model().objects.all():
+                    connection.set_tenant(tenant)
```
eodolphi: This should probably reset the tenant. I think we have a context manager for that.

bluebottle/journals/tests/test_admin.py
```diff
@@ -0,0 +1,39 @@
+# from bluebottle.test.utils import BluebottleTestCase
```
eodolphi: this should probably be removed.

setup.py
```diff
@@ -75,19 +75,20 @@ def read_file(name):
...
     'fabric',
-    'django-tenant-schemas==1.5.2',
+    'django-tenant-schemas==1.5.7',
     'raven==5.1.1',
     'regex==2015.11.22',
     'djangorestframework-jwt==1.7.2',
     'django-filetransfers==0.1.0',
     'django-admin-tools==0.5.2',
     'django-rest-swagger==0.3.4',
+    'django-webtest==1.7.8',
```
eodolphi: This should probably only be in tests_requires. If not, the version should match.

tests.txt
eodolphi: this should probably be removed.

bluebottle/accounting/utils.py
gannetson: Clueless on this one...

eodolphi: Yeah me too. It seems to be a dict that can where sub dicts are added too. I think:

```
mydict(a=mydict(b='c'), b=3) + mydict(a=mydict(d='e'), c=4) == {'a': {'b': 'c', 'd': 'e'}, 'b': 3, 'c': 4}
``` 

Or something like that. Maybe we can copy some examples from the tests?

bluebottle/accounting/views.py
gannetson: Not sure how to do that. Is this not resetting the tenant cleanly?
This code does work, I think.

eodolphi: I think it does work, but sort of by accident. After the connection.set_tenant, there seems to be no database interaction. If there would be that would happen on the last tenant in the loop. The following request, the connection is set correctly again.

I think
```
with LocalTenant(tenant):
    ....
```

Would be the correct way.

bluebottle/accounting/utils.py
gannetson: @sergei-maertens could you shine some light on this?

sergei-maertens: I have forwarded it to Jos who wrote this initially, I'm in a meeting all
day.

gannetson: @josvromans

josvromans: Dictionary with a custom add method that uses the 'add_up' function to handle all kinds of values (another dict, a list, value, string, None, ..)

When dictA (self) is an empty dict, return dictB (without lookint at the structure)

Only add up two dicts when both have exactly the same keys (return empyty MyDict when they do not have the same keys)

Construct a new MyDict, with exactly the same keys as the two dicts, and with values that are calculated by the add_up function that can handle any kind of nested dicts, lists (of dicts) etc. (this same add_up function is used when each item in two MyLists are added up)

bluebottle/accounting/utils.py
josvromans: List with a custom add method that uses the 'add_up' function to handle all kinds of values (another dict, a list, value, string, None, ..)

Return an empty mylist when the two given lists do not have the same length.

When both lists have the same length, construct a new MyList, with the same length as the two given lists, and with values that are calculated by the add_up function that can handle any kind of nested dicts, lists (of dicts) etc. (this same add_up function is used when each value in two MyDicts are added up)

PR #1823, opened 2015-11-20
Order export projects above delete projects

eodolphi:


bluebottle/projects/admin.py
```diff
@@ -168,6 +169,11 @@ def get_list_editable(self, request):
 
     actions = (export_as_csv_action(fields=export_fields), )
 
+    def get_actions(self, request):
+        "Order the action in reverse (delete at the bottom)."
```
gannetson: I think like this it is a string, not a comment.

PR #1847, opened 2015-12-09
Feauture/changes joris

aethemba:


bluebottle/projects/models.py
```diff
@@ -425,6 +428,25 @@ def status_changed(self, old_status, new_status):
...
+            "Project": self.title,
+            "Owner": self.owner.email,
+        }
+        mp = None
+        KEY = getattr(properties, 'MIXPANEL', None)
+
+        if KEY:
+            mp = Mixpanel(KEY)
+
+        if mp and old_status.sequence in (1, 2, 3, 4, 5, 6) and new_status.sequence in (8, 9, 10):
```
rollick: Should we use the slug here rather than sequence?

aethemba:
@rollick , good idea. I made the necessary changes.

PR #1849, opened 2015-12-16
Add setting for read only profile fields BB-5316

rollick:


bluebottle/settings/base.py
```diff
@@ -290,6 +290,8 @@
     'django.core.context_processors.static',
     'django.core.context_processors.tz',
     'django.contrib.messages.context_processors.messages',
+    'tenant_extras.context_processors.conf_settings',
```
eodolphi: I cannot find that context processor anywhere

bluebottle/utils/context_processors.py
```diff
@@ -1,8 +1,33 @@
...
+
+
+def tenant_properties(request):
+    extra_properties = tenant_extra_properties(request)
+    settings = json.loads(extra_properties['settings'])
+
+    # If tenant has SAML enabled then we also return a list
+    # of read-only user profile properties.
+    try:
+        mappings = properties.TOKEN_AUTH['assertion_mapping']
```
eodolphi: Maybe the name should indicate that this is to set the read only properties.

Can't this be decoupled slightly more from tenant_extra_properties a little by this just doing the read only setting?

rollick: It is extending the "settings" property in the context. I wanted to just move the context processor out of tenant-extras packages but thought we can do that later.

bluebottle/settings/base.py
rollick: It is in the tenant-extras package.

bluebottle/utils/context_processors.py
eodolphi: Yeah I saw that it was extending it, I just thought that we could decouple this a little more. If we'll move the context processor out, then my comment above is not relevant.

bluebottle/settings/base.py
eodolphi: I mis-typed. Never mind.

PR #1850, opened 2015-12-16
Ensure read-only user profile fields are not editable via api BB-5316

rollick:


bluebottle/bb_accounts/views.py
```diff
@@ -64,6 +65,22 @@ class ManageProfileDetail(generics.RetrieveUpdateAPIView):
     permission_classes = (TenantConditionalOpenClose, IsCurrentUser)
     serializer_class = ManageProfileSerializer
 
+    def pre_save(self, obj):
+        try:
+            # Read only properties come from the TOKEN_AUTH / SAML settings
+            assertion_mapping = properties.TOKEN_AUTH['assertion_mapping']
```
eodolphi: Why doesn't this use the readonly fields settings from the previous pull request?

rollick: @ernst the settings from the previous PR is a json string returned in a context processor. The only bits in common I think are:
```python
assertion_mapping = properties.TOKEN_AUTH['assertion_mapping']
user_properties = assertion_mapping.keys()
```

PR #1865, opened 2016-01-18
Fix status changing of payouts

eodolphi:


gannetson:
Not sure if Nguyen is happy with this, that you can't change status in a detail view.
We'll hear about this later I guess. We might have to change it later and just make the status field editable (non-protected) later after all.

eodolphi:
Yeah, that was my first guess too. However this was the way it was supposed to work in the first place, and setting the status in the detail view did not go through the FSM methods. That can probably be fixed, but I did not see why immediately.

PR #1871, opened 2016-02-01
Change Statistics model

gannetson:


bluebottle/statistics/migrations/0005_auto__add_field_statistic_language__add_field_statistic_title_nl__add_.py
```diff
@@ -0,0 +1,55 @@
...
+        # Adding field 'Statistic.title_nl'
+        db.add_column(u'statistics_statistic', 'title_nl',
+                      self.gf('django.db.models.fields.CharField')(max_length=100, null=True, blank=True),
+                      keep_default=False)
+
+        # Adding field 'Statistic.title_en'
+        db.add_column(u'statistics_statistic', 'title_en',
+                      self.gf('django.db.models.fields.CharField')(max_length=100, null=True, blank=True),
+                      keep_default=False)
+
```
eodolphi: This does not work correctly. It creates the dutch and english version for all tenants. How do we add a thirds language only for a specific client?

eodolphi: Would it be hard to create a version of  ModelTranslation that looks at the properties instead of the settings?

One thing we could do is change this migration to use the actual properties and a correct columns accordingly. Changing a sides language will still fail however.

PR #1880, opened 2016-02-09
Feature/fix stats translations

gannetson:


bluebottle/homepage/views.py
```diff
@@ -9,6 +11,11 @@ class HomePageDetail(generics.GenericAPIView):
     serializer_class = HomePageSerializer
 
     def get(self, request, language='en'):
+
+        # Force requested language
+        translation.activate(language)
```
eodolphi: Don't we have a context manager for this? Something like ```with TenantLanguage(language)``` or something...

gannetson: I tried to find a context manager for this, but this url is rather different from others. We specifically ask for a language, that's why I thought it was better to set it here and like this.
What would you suggest @eodolphi ?

PR #1900, opened 2016-02-22
Feature/bb 5964

aethemba:


bluebottle/wallposts/models.py
```diff
@@ -72,6 +72,11 @@ class Wallpost(PolymorphicModel):
     share_with_linkedin = models.BooleanField(default=False)
     email_followers = models.BooleanField(default=True)
 
+    donation = models.ForeignKey('donations.Donation',
+                                 verbose_name=_("Donation"),
+                                 related_name='donation',
+                                 null=True, blank=True)
+
```
gannetson: Have you thought about using content_object to point that to the donation?

aethemba: Ah no, I did not think of that. Interesting, but this sets the donation on the wallpost, right? Not the other way around, or wouldn't that matter?

PR #1919, opened 2016-03-11
Load the right settings file for tenant

gannetson:


bluebottle/clients/middleware.py
```diff
@@ -19,17 +19,13 @@ def set_tenant(self, tenant):
...
-                props_mod = safe_join(settings.MULTI_TENANT_DIR,
-                                      tenant.client_name,
-                                      "{0}.py".format(settings_file))
-                # try to load tenant specific properties. We're using execfile since tenant
-                # directories are not python packages (e.g. no __init__.py)
-                execfile(props_mod, dict(settings=settings),
-                         self.tenant_properties)
-        except (ImportError, AttributeError, IOError):
-            pass
+        props_mod = safe_join(settings.MULTI_TENANT_DIR,
```
rollick: Why did you remove the try/except?

gannetson: It should just be there I think. It's better to get a blunt error than it to proceed without notice.

PR #1924, opened 2016-03-14
Small clean up

rollick:


bluebottle/common/templates/base.mail.html
```diff
@@ -32,7 +32,7 @@
                                                     <tr>
                                                         <td class="twelve columns">
 
-                                                            <img src="{{site}}{{TENANT_MAIL_PROPERTIES.logo}}">
+                                                            <img src="{{site}}/images/logo-email.gif">
```
gannetson: Why not use the path pfrom settings?

rollick: It is the same for all tenants.

PR #1977, opened 2016-05-10
Feature/bb 6820 remove get serializer class

gannetson:


bluebottle/bb_accounts/views.py
```diff
@@ -22,7 +22,7 @@
 from bluebottle.clients import properties
 from bluebottle.members.serializers import (
     UserCreateSerializer, ManageProfileSerializer, UserProfileSerializer,
-    PasswordResetSerializer, PasswordSetSerializer)
+    PasswordResetSerializer, PasswordSetSerializer, CurrentUserSerializer)
 
 USER_MODEL = get_user_model()
```
rollick: Is this going to be removed in a different branch?

bluebottle/utils/serializers.py
```diff
@@ -282,27 +282,27 @@ def _get_callable(self, obj, attr):
         return None
 
 
-class DefaultSerializerMixin(object):
-    def get_custom_serializer(self):
-        return self.model._meta.default_serializer
+# class DefaultSerializerMixin(object):
```
rollick: Can we just remove this? We can always find it again in a commit if needed.

PR #1992, opened 2016-05-21
Bump jwt package to 1.7.2

rollick:


gannetson:
Ah so this is the latest passing (Django 1.6.8 compatibel) version of drf-jwt?

PR #2032, opened 2016-05-27
Feature/django upgrade projects

gannetson:


bluebottle/bb_projects/permissions.py
```diff
@@ -50,7 +50,8 @@ def _get_project_from_request(self, request):
             try:
                 project = Project.objects.get(slug=project_slug)
                 return project
-            except Project.DoesNotExist:
+            except Project.\
```
eodolphi: Why is this?

PR #2060, opened 2016-06-06
dont show pledge in admin if its not a payment option

aethemba:


bluebottle/members/admin.py
```diff
@@ -111,18 +112,38 @@ class MemberVotesInline(admin.TabularInline):
...
+                        'first_name', 'last_name', 'username', 'gender', 'birthdate',
+                        'phone_number')}),
+                    (_("Profile"),
+                     {'fields': ('user_type', 'is_co_financer', 'can_pledge', 'picture', 'about_me', 'location')}),
+                    (_("Settings"), {'fields': ['primary_language', 'newsletter']}),
+                    (_('Skills & interests'), {'fields': ('favourite_themes',)}),
+                    (_('Important dates'),
+                     {'fields': ('last_login', 'date_joined', 'deleted')}),
+                )
+                break
```
eodolphi: Why don't you add the field to the configuration instead of copying everything over? 

I would also just remove the break. Looping over 2 more payment methods will not kill performance.

bluebottle/members/admin.py
```diff
@@ -111,18 +112,38 @@ class MemberVotesInline(admin.TabularInline):
...
-            'first_name', 'last_name', 'username', 'gender', 'birthdate',
-            'phone_number')}),
-        (_("Profile"),
-         {'fields': ('user_type', 'is_co_financer', 'can_pledge', 'picture', 'about_me', 'location')}),
-        (_("Settings"), {'fields': ['primary_language', 'newsletter']}),
-        (_('Skills & interests'), {'fields': ('favourite_themes',)}),
-        (_('Important dates'),
-         {'fields': ('last_login', 'date_joined', 'deleted')}),
-    )
+    def get_standard_fieldsets(self):
```
eodolphi: If you make this a property, there is no need for the changes to `change_view`.

PR #2064, opened 2016-06-07
Merge develop into django upgrade

gannetson:


rollick:
Done is a different PR

PR #2087, opened 2016-06-14
Feature/booking openup backend

aethemba:


bluebottle/suggestions/views.py
```diff
@@ -29,13 +29,13 @@ def get_queryset(self):
 
 class SuggestionDetail(generics.RetrieveUpdateAPIView):
     queryset = Suggestion.objects.all()
-    permission_classes = (IsAuthenticated,)
+    permission_classes = (AllowAny,)
```
eodolphi: Is this a potential security issue? Anybody can read update and delete suggestions. That does not seem right.

eodolphi: It seems you can even view the token in the detail view. So yes, this seems like an security issue.

bluebottle/suggestions/views.py
```diff
@@ -1,14 +1,14 @@
...
 
 from bluebottle.suggestions.models import Suggestion
 from bluebottle.suggestions.serializers import SuggestionSerializer
 
 
 class SuggestionList(generics.ListCreateAPIView):
     queryset = Suggestion.objects.all()
-    permission_classes = (IsAuthenticated,)
+    permission_classes = (AllowAny,)
     serializer_class = SuggestionSerializer
```
eodolphi: Same here, not anybody should be able to read the list of suggestions. Anybody should be able to create them though.

bluebottle/utils/email_backend.py
```diff
@@ -97,11 +97,17 @@ def _send(self, email_message):
...
                    from_email=None, **kwargs):
+
     if hasattr(to, 'primary_language') and to.primary_language:
         language = to.primary_language
     else:
         language = properties.LANGUAGE_CODE
 
+    # This is an exception to handle a Bookingcares.com language which
+    # contains more languages than the rest of the platform
+    if kwargs.get('language', None):
```
eodolphi: `if 'language' in kwargs:` looks more pyhtonic to me.

bluebottle/settings/base.py
```diff
@@ -145,6 +145,7 @@
     'bluebottle.auth.middleware.AdminOnlySessionMiddleware',
     'bluebottle.auth.middleware.AdminOnlyCsrf',
     'bluebottle.auth.middleware.AdminOnlyAuthenticationMiddleware',
+    'corsheaders.middleware.CorsMiddleware',
```
eodolphi: This also requires a change to the settings generated in server-deployment right?

bluebottle/suggestions/views.py
aethemba: In still working on this, its not meant to be merged yet. Ill add Do Not Merge to the title

PR #2102, opened 2016-06-21
Feature/bb 7146 fix payouts

gannetson:


rollick:
We should merge master into develop to get these changes.

PR #2113, opened 2016-06-22
Add Remote ID to exports

gannetson:


eodolphi:
Cool! But I do not think this will solve the complete issue. I think the idea was to add it to the normal csv exports.

PR #2139, opened 2016-07-01
Fix scaling of profile images

gannetson:
BB-6967 #resolve

### DIFF
--- a/bluebottle/bluebottle_drf2/serializers.py
+++ b/bluebottle/bluebottle_drf2/serializers.py
@@ -39,11 +39,10 @@ class RestrictedImageField(serializers.ImageField):
 
 
 class SorlImageField(RestrictedImageField):
-    def __init__(self, geometry_string, **kwargs):
+    def __init__(self, geometry_string, crop='center', colorspace='RGB', **kwargs):
         self.geometry_string = geometry_string
-        self.kwargs = kwargs
-        self.crop = kwargs.pop('crop', 'center')
-        self.colorspace = kwargs.pop('colorspace', 'RGB')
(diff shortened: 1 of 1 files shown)


bluebottle/bluebottle_drf2/serializers.py
```diff
@@ -40,10 +40,10 @@ def to_internal_value(self, data):
 
 class SorlImageField(RestrictedImageField):
     def __init__(self, geometry_string, **kwargs):
-        self.crop = kwargs.pop('crop', 'center')
```
eodolphi: Since we do not use kwargs anymore, wouldn't it be cleaner to just change the signature here to:

```
def __init__(self, geometry_string, crop='center', colorspace='RGB'):
```

gannetson: The idea is that we could pass kwargs like here:
https://github.com/onepercentclub/bluebottle/blob/c87da0d0b79c55e3f79cb98bcfd9129c6be5bbef/bluebottle/members/serializers.py#L42

eodolphi: That will still work. It just makes it clear what the signature of the function is. And the code that sets the defaults is now gone.

aethemba:
@gannetson , @eodolphi , is this ready to go in?

bluebottle/bluebottle_drf2/serializers.py
gannetson: Yeah true. The original idea was that you could pass other imagick properties, but since we don't pass them  we could skip using this.kwargs

PR #2147, opened 2016-07-07
added task filter on TaskList view and added tests

aethemba:


bluebottle/bb_tasks/views.py
```diff
@@ -79,6 +83,21 @@ def get_queryset(self):
...
+        end_date = self.request.query_params.get('end_date', None)
+
+        # User searches for tasks on a specific day.
+        if start_date and not end_date:
+            qs = qs.filter(Q(type='event', deadline=start_date) |
+                           Q(type='ongoing', deadline__gte=start_date))
+
+        # User searches for tasks in a specific range
+        if start_date and end_date:
+            qs = qs.filter(deadline__range=[start_date, end_date])
```
gannetson: ongoing tasks with deadline > end_date should also be returned I think (because they are 'active' within the range)

aethemba: Which line are you referring to? Because line 99 doesn't filter on type and if there is an explicit end_date it shouldn't return anything outside of that range.

aethemba: @gannetson , ^^

PR #2188, opened 2016-08-09
Validate that currencies match the project currency

eodolphi:


eodolphi:
This fails because multipart api calls do not work with our currency objects: https://onepercentclub.atlassian.net/browse/BB-7617

gannetson:
We now accept flattened MoneyFields to (json string) @eodolphi

eodolphi:
Jep I'll fix the tests.

PR #2189, opened 2016-08-09
Facebook login without email

rollick:


rollick:
@eodolphi We should get this merged asap. The frontend already has some handling for invalid email addresses when the user goes to a project page.

PR #2220, opened 2016-08-23
Lazy currency loading. Add currencies to cfg api.

gannetson:
+ Read currencies from tenant settings (lazily).
+ Expose currencies in config api.

BB-7633 #resolve
BB-7634 #resolve

### DIFF
--- a/bluebottle/settings/base.py
+++ b/bluebottle/settings/base.py
@@ -532,6 +532,14 @@ PROJECT_PAYOUT_FEES = {
     'not_fully_funded': .05
 }
 
+CURRENCIES_ENABLED = [
+    {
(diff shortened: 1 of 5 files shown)


bluebottle/settings/base.py
```diff
@@ -532,6 +532,14 @@
...
 }
 
+CURRENCIES_ENABLED = [
+    {
+        'code': 'EUR',
+        'name': 'Euro',
+        'symbol': u"\u20AC"
+    }
+]
+
```
gannetson: @ernst I took your suggestion in consideration, but in the end I think this more verbose method is slightly better than derive it from payment methods. Are you ok with this?

PR #2241, opened 2016-08-31
Fix saving users with a location but not address

eodolphi:


eodolphi:
Without the changes the new tests raises:

RelatedObjectDoesNotExist: Member has no address.


Just like we see on the server

PR #2255, opened 2016-09-06
Fix for "extra" group permissions not loading on migrate

rollick:
BB-7830 #resolve

Plus some pep8 fixes.


rollick:
On migrate we should have been loading some extra permissions but this was being skipped:
https://github.com/onepercentclub/bluebottle/blob/develop/bluebottle/utils/models.py#L53-L76

PR #2262, opened 2016-09-12
Feature/interswitch

gannetson:


bluebottle/payments_interswitch/tests/test_api.py
```diff
@@ -0,0 +1,164 @@
+import unittest
```
eodolphi: It looks to me that this does not test the interwswitch adapter.

Please replace this with tests for the interswitch module.

gannetson: Bad copy-paste job :-(

PR #2277, opened 2016-09-19
Add management command + tweaks

gannetson:


bluebottle/surveys/adapters.py
```diff
@@ -14,31 +17,93 @@ def update_surveys(self):
...
+            if match:
+                question = match.group(1)
+                if answers.has_key(question):
+                    answers[question] += ", " + data[key]
+                else:
+                    answers[question] = data[key]
+        return answers
+
+    def get_responses(self, survey):
+        self.client.config.response_type = 'json'
```
eodolphi: Why do we only use json here?

gannetson: For `list` calls it fails :-( only for `get`.
Guess we should fix the library

PR #2283, opened 2016-09-21
Feature/improve survey api

gannetson:


bluebottle/surveys/urls/api.py
```diff
@@ -4,5 +4,5 @@
 
 urlpatterns = patterns(
     '',
-    url(r'^(?P<slug>[\w-]+)/$', ProjectSurveyList.as_view(), name='project_survey_list'),
+    url(r'^$', ProjectSurveyList.as_view(), name='project_survey_list'),
```
eodolphi: Is ember incompatible with this way of specifying the url? 

Required query arguments feel a little wrong to me....

gannetson: We could compose the url like that I guess, but we need to write more code in the adapters.
Maybe new Ember supports the nested urls. For now it's better to have a surveys end point and 'filter' on project.

PR #2290, opened 2016-09-23
Start of analytics app for exporting model metrics

rollick:
* added basic tags gathering for projects, tasks and task members BB-7932
* add basic analytics integration with influxdb as backend BB-7932

### DIFF
--- a/bluebottle/analytics/backends.py
+++ b/bluebottle/analytics/backends.py
@@ -1,0 +1,31 @@
+from influxdb import InfluxDBClient
+
+from django.conf import settings
+
+
+class InfluxExporter:
+    def __init__(self, conf):
+        self.measurement = conf['measurement']
+        self.domain = conf['domain']
+        self.port = conf['port']
+        self.username = conf['username']
+        self.password = conf['password']
+        self.database = conf['database']
+
+        self.client = self._client
+
+    def _client(self):
+        return InfluxDBClient(self.domain, self.port, self.username,
+                              self.password, self.database)
+
+    def process(self, timestamp, tags={}, fields={}):
+        json_body = [
+            {
+                "measurement": self.measurement,
+                "time": timestamp ,
+                "tags": tags,
+                "fields": fields
+            }
+        ]
+
+        self.client.write_points(json_body)
--- a/bluebottle/analytics/signals.py
+++ b/bluebottle/analytics/signals.py
@@ -1,5 +1,7 @@
 import re
 
+from django.utils import timezone
+from django.conf import settings
 from django.utils.translation import ugettext as _
 from django.db.models.signals import post_save
 from django.dispatch import receiver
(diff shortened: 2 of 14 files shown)


bluebottle/analytics/backends.py
```diff
@@ -0,0 +1,31 @@
...
+class InfluxExporter:
+    def __init__(self, conf):
+        self.measurement = conf['measurement']
+        self.domain = conf['domain']
+        self.port = conf['port']
+        self.username = conf['username']
+        self.password = conf['password']
+        self.database = conf['database']
+
+        self.client = self._client
```
eodolphi: shouldn't this be: 
```self.client = self._client()```

Or make _client a property?

File: bluebottle/analytics/signals.py [hunk 2]
```diff
@@ -0,0 +1,91 @@
...
+from .utils import queue_analytics_record
+
+
+@receiver(post_save, weak=False, dispatch_uid='model_analytics')
+def post_save_analytics(sender, instance, **kwargs):
+    if not getattr(settings, 'ANALYTICS_ENABLED', False):
+        return
+
+    # Return early if instance is a migration.
+    if instance.__class__.__name__ == 'Migration':
```
eodolphi: If we skip all migration, doesn't that mean the metrics become out of sync?

bluebottle/analytics/utils.py
```diff
@@ -0,0 +1,32 @@
+from importlib import import_module
+
+from django.conf import settings
+
+
+# NOTE: Stubbed method for testing
+def queue_analytics_record(timestamp, tags={}, fields={}):
+    # Get the default backend for analytics
+    def get_handler_class(handler):
```
eodolphi: Why is this function defined inline? Do we need a new one each time this function is called? Or does it need to be private?

bluebottle/analytics/signals.py
rollick: What part of the metrics could be out of sync? This is only ignoring signals from the Migration class as these classes never generate any metrics data. They would exit later anyway when checking further down for instance.Analytics but I thought it is more effiicient to just exit early.

bluebottle/analytics/utils.py
rollick: It isn't necessary now but I should be looping over all possible backends as there can be multiple. This will be added later and then I think it is neater to have an inline method.

I should really move it into the utils app as something similar is being used in other parts of the code to get a class from a dotted string.

bluebottle/analytics/backends.py
rollick: Fixed.

bluebottle/orders/models.py
```diff
@@ -1,8 +1,22 @@
 from bluebottle.bb_orders.models import BaseOrder
+from bluebottle.utils.utils import PreviousStatusMixin
 
 
-class Order(BaseOrder):
-    pass
+class Order(BaseOrder, PreviousStatusMixin):
+    @property
+    def anonymous(self):
+        return True if self.user else False
```
gannetson: Is this the right way around?

bluebottle/projects/models.py
```diff
@@ -515,6 +515,18 @@ def get_tweet(self, **kwargs):
...
+    class Analytics:
+        type = 'project'
+        tags = {
+            'sub_type': 'project_type',
+            'status': 'status.name',
+            'status_slug': 'status.slug',
+            'theme': 'theme.name',
+            'theme_slug': 'theme.slug',
+            'location': 'location.name',
+            'country': 'location.country.name'
```
gannetson: If the platform doesn't use locations the country is set directly on the project (`project.country`).
Furthermore you might get a property error with `location.name` if there is no location set.

bluebottle/orders/models.py
rollick: Nice catch.

bluebottle/projects/models.py
rollick: I will update the country v. location.

The analytics part catches attribute errors so no location.name is ok.

bluebottle/analytics/signals.py
eodolphi: I was thinking about data migration, that moved the status of a whole bunch of things for example. We have had these migrations before.

eodolphi: But maybe I just do not really understand when the class has the name 'Migration'.

PR #2295, opened 2016-09-26
Feature/survey task project aggregate

gannetson:


bluebottle/surveys/adapters.py
```diff
@@ -45,7 +45,8 @@ def parse_answers(self, data):
             match = question_re.match(key)
             if match:
                 question = match.group(1)
-                answers[question] = data[key]
+                if bool(data[key]):
```
eodolphi: Why cast to bool? Does'nt 
```
if data[key]:
```
work?

bluebottle/surveys/models.py
```diff
@@ -101,6 +224,7 @@ class Response(models.Model):
     submitted = models.DateTimeField(null=True, blank=True)
     project = models.ForeignKey('projects.Project', null=True, blank=True)
     task = models.ForeignKey('tasks.Task', null=True, blank=True)
+    user_type = models.CharField(max_length=200, blank=True, null=True)
```
eodolphi: maybe we can be explicit about the possible values: (task_member, initiator, organization).

bluebottle/surveys/adapters.py
gannetson: if {} => True
if bool({}) => False

bluebottle/surveys/models.py
gannetson: True

PR #2301, opened 2016-09-27
Feature/survey retrieve paged results

eodolphi:


bluebottle/surveys/adapters.py
```diff
@@ -185,7 +210,9 @@ def update_survey(self, survey):
                     except SubQuestion.DoesNotExist:
                         pass
 
+        survey.last_synced = timezone.now()
```
gannetson: Are you sure they use timezone too? We don't want to mis two hours of surveys..

eodolphi: It should not matter. For one thing, we use the same date again when querying. This cancels out any offsets. And second, we use only the day of this date, and then ask for everything with this day or later, (doing some double work, but being more precise is not possible in their api).

PR #2305, opened 2016-09-27
Don't use actual images for tests (breaks reef)

gannetson:


gannetson:
Cherry picked from integration/multicurrency

PR #2330, opened 2016-09-30
Add test for changing tasks member status BB-8115

rollick:


bluebottle/analytics/tests/test_models.py
```diff
@@ -220,7 +233,6 @@ def test_tags_generation(self, queue_mock):
             'id': order.id
         }
 
-        print queue_mock.call_args_list
```
gannetson: Did you write a test to cover up this print removal? :-P

PR #2333, opened 2016-10-03
Add analytics for wallpost reactions

eodolphi:
BB-8138 #resolve

### DIFF
--- a/bluebottle/wallposts/models.py
+++ b/bluebottle/wallposts/models.py
@@ -209,6 +209,22 @@ class Reaction(models.Model):
     objects = ReactionManager()
     objects_with_deleted = models.Manager()
 
+    class Analytics:
+        type = 'wallpost'
+        tags = {}
+        fields = {
+            'id': 'id',
+            'user_id': 'author.id'
+        }
+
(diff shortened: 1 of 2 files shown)


bluebottle/wallposts/models.py
```diff
@@ -209,6 +209,22 @@ class Reaction(models.Model):
...
 
+    class Analytics:
+        type = 'wallpost'
+        tags = {}
+        fields = {
+            'id': 'id',
+            'user_id': 'author.id'
+        }
+
+        def extra_tags(self, instance, created):
```
rollick: You can put this in the list of tags in the class as it won't change based on the instance.

bluebottle/wallposts/models.py
```diff
@@ -209,6 +209,22 @@ class Reaction(models.Model):
...
+            return {'sub_type': 'reaction'}
+
+        def skip(self, obj, created):
```
rollick: Probably don't need this either. The reactions are always from users so even if the reaction is to  system wallpost we should record the event.

PR #2367, opened 2016-10-14
Hotfix/log auth failures

eodolphi:


bluebottle/members/serializers.py
```diff
@@ -47,7 +47,7 @@ def __init__(self, *args, **kwargs):
 
     class Meta:
         model = BB_USER_MODEL
-        fields = ('id', 'first_name', 'last_name', 'username',
```
gannetson: In some parts of Reef we still rely on usernames, e.g.
https://github.com/onepercentclub/reef/blob/master/frontend/app/controllers/task.js#L44-L54
We should fix that before we deploy this.

eodolphi: That is already fixed. 

I do not really understand why this is part of the pull request. I already merged these commits to the release branch...

PR #2384, opened 2016-10-19
Feature/currency symbols in templates

eodolphi:


gannetson:
Been a while since we did django template tags :-)

PR #2398, opened 2016-10-25
Add project location / region to Taskmember export. Add survey to pro…

eodolphi:
…ject

export.

BB-8229 #resolve


bluebottle/projects/models.py
```diff
@@ -423,6 +423,10 @@ def is_realised(self):
     def is_funding(self):
         return self.amount_asked.amount > 0
 
+    @property
+    def has_survey(self):
+        return len(self.response_set.all()) > 0
```
gannetson: self.responses_set.count() > 0

eodolphi: That is the same right?

PR #2399, opened 2016-10-25
Make sure that a mail is send when user withdraws from task.

eodolphi:
Use the user's full name in the body

BB-8242 #resolve
BB-8140 #resolve

### DIFF
--- a/bluebottle/tasks/tests/test_mails.py
+++ b/bluebottle/tasks/tests/test_mails.py
@@ -42,8 +42,7 @@ class TestTaskMemberMail(TaskMailTestBase):
         self.assertEquals(len(mail.outbox), 1)
         body = mail.outbox[0].body
         self.assertTrue('applied for your task' in body)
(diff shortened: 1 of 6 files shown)


bluebottle/tasks/tests/test_mails.py
```diff
@@ -40,8 +40,10 @@ def test_member_applied_to_task_mail(self):
 
         # Task owner receives email about new task member
         self.assertEquals(len(mail.outbox), 1)
-        self.assertNotEquals(mail.outbox[0].body.find("applied for your task"),
-                             -1)
+        body = mail.outbox[0].body
+        self.assertTrue('applied for your task' in body)
+        self.assertTrue(self.task_member.member.first_name in body)
+        self.assertTrue(self.task_member.member.last_name in body)
```
gannetson: I think the BluebottleUserFactory creates objects with first_name=user_1 and last_name=user_1, these test might not prove that you're not using short_name.

PR #2408, opened 2016-10-26
Fix vitepay payments

eodolphi:
Note that for development we use a callback url that always returns
true.

For testing/staging we open up the callback url in django-lockdown.

BB-8348


rollick:
@ernst this looks like develop into release-19.

eodolphi:
Yeah, I know. Loek pushed 15 commits to this branch. I do not think they
belong here. I'll sort it out tomorrow morning.

Op 27 okt. 2016 11:52 schreef "rollick" <notifications@github.com>:

@ernst <https://github.com/ernst> this looks like develop into release-19.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/onepercentclub/bluebottle/pull/2408#issuecomment-256598133>,
or mute the thread

PR #2423, opened 2016-11-03
Make sure project is succesful if amount extra + donated > asked

gannetson:
BB-8418 #resolve


eodolphi:
Did this need to go in the release?

If so, can you recreate it against master?

gannetson:
I think thos has been broken for a while, but let's create it against master en start figuring out why it's failing...

PR #2456, opened 2016-11-16
Feature/title in project payout

eodolphi:


bluebottle/payouts_dorado/serializers.py
```diff
@@ -23,23 +28,25 @@ class Meta:
...
 
-    amount_asked = MoneySerializer()
-    amount_donated = MoneyTotalSerializer(source='totals_donated')
-
-    receiver_account_name = serializers.CharField(source='account_holder_name')
-    receiver_account_number = serializers.CharField(source='account_number')
-    receiver_account_bic = serializers.CharField(source='account_bic')
-    receiver_account_city = serializers.CharField(source='account_holder_city')
-    receiver_account_country = serializers.CharField(source='account_holder_country.name')
+    title = serializers.CharField(required=False)
```
gannetson: Why don't we use read_only=True instead of required? Is there a case where we do want to write these?

eodolphi: You are right. Fixed!

PR #2481, opened 2016-12-01
Check correct celery setting for monthly donations

rollick:
And avoid double negative when checking celery in analytics


gannetson:
Thanks!

PR #2500, opened 2016-12-12
Feature/platform results translations

eodolphi:


bluebottle/utils/monkey_patch_parler.py
```diff
@@ -0,0 +1,29 @@
+import parler.appsettings
+from parler.utils.conf import add_default_language_settings
+
+from bluebottle.clients import properties
+
+
+appsettings = parler.appsettings
+
+
+class TenantAwareParlerAppsettings(object):
```
rollick: Are there any tests for this? I presume the translations will only be seen in the admin site so "properties" will be set correctly but otherwise there could be problems.

eodolphi: Good call, I added tests for this now.

PR #2518, opened 2016-12-16
Feature/result pages data migrations

eodolphi:


bluebottle/cms/models.py
```diff
@@ -125,7 +125,7 @@ def __unicode__(self):
 
 class ProjectsContent(ResultsContent):
     action_text = models.CharField(max_length=40,
-                                   default=_('Add your own project'),
+                                   default=_('Startyour own project'),
```
gannetson: Sure?

eodolphi: Fixed...

PR #2554, opened 2017-02-07
Feature/flutterwave

gannetson:


bluebottle/payments_flutterwave/models.py
```diff
@@ -0,0 +1,107 @@
...
+    auth_model = models.CharField(
+        help_text="Authentication Model - BVN, PIN, NOAUTH, VBVSECURECODE",
+        default="BVN",
+        null=True, blank=True,
+        max_length=200)
+    card_number = models.CharField(
+        help_text="Card Number",
+        null=True, blank=True,
+        max_length=200)
+    cvv = models.CharField(
```
ashutoshb: should we really store all the credit card information, including the CVV in plain text ? can we make a oneway hash of it ?

gannetson: Yeah you're right. We need to send it to Fluttwerwave API though. Do you know a way to make that work?

ashutoshb: I haven't gone thru the whole api payment flow and dont know if the request to the api payment gateway is made synchronously during the request or asynchronously. I will go thru whole flow and also check up if there is any way of doing a one way hash of the information.

bluebottle/payments_flutterwave/models.py
```diff
@@ -0,0 +1,107 @@
...
+        default="NG",
+        null=True, blank=True,
+        max_length=200)
+
+    transaction_reference = models.CharField(
+        help_text="Flutterwave transaction reference",
+        null=True, blank=True,
+        max_length=200)
+
+    response = models.CharField(help_text=_('Response from Flutterwave'),
```
ashutoshb: can we make this a TextField to avoid truncation of the JSON response if it is greater than 5000 chars ? This also makes sure that we always store the full response from flutterwave and nothing gets lost.

bluebottle/payments_flutterwave/models.py
```diff
@@ -0,0 +1,107 @@
...
+        max_length=100)
+    narration = models.CharField(
+        help_text="Transaction description",
+        null=True, blank=True,
+        max_length=200)
+    response_url = models.CharField(
+        help_text="Callback Url",
+        null=True, blank=True,
+        max_length=200)
+    country = models.CharField(
```
ashutoshb: since flutterwave only supports NG, GH, UK, US, KE, [Flutterwave API](https://www.flutterwave.com/documentation/card-payments/#validate), can we set choices on the field ? This would help us harden the validation.

gannetson: They are looking at expanding. Now we always set it to 'NG'. 
Don't think this is blocker for now, right? @ashutoshb

bluebottle/payments_flutterwave/models.py
gannetson: Yeah, I already upped it from 1000 to 5000. 
I will make it so.

PR #2560, opened 2017-02-13
Check auto deploy

gannetson:


ashutoshb:
Do we also need to remove the quotes on line 15, $TRAVIS_PULL_REQUEST == 'false' ?

gannetson:
I think the quotes are optional either way... It was working Friday for testing and develop deploys.

PR #2565, opened 2017-02-17
BB-8754 weekly project status statistics

ashutoshb:


eodolphi:
Looks good. 

2 remarks:

- The update_status_stats is not tested.
- Maybe we should run this daily. All other data is live, and the difference between live data and week old data might be a bit big.

ashutoshb:
Thanks for the suggestions. I'll update the PR with the changes.

eodolphi:
Cool!

PR #2570, opened 2017-02-21
Test that we actually set that payout status to needs_approval when a

eodolphi:
project finishes

BB-9043 #resolve


ashutoshb:
Looks good. 👍

PR #2571, opened 2017-02-21
ensure task deadline is set to midnight server time when updating task BB-9022

rollick:


eodolphi:
This looks good, but I'll wait with merging until after the flutterwave stuff settles.

PR #2584, opened 2017-02-27
Order the projects in the results page map block on campaign ended, so

eodolphi:
taht it is possible to display the most recent project

BB-8831 #resolve

### DIFF
--- a/bluebottle/cms/serializers.py
+++ b/bluebottle/cms/serializers.py
@@ -130,7 +130,10 @@ class ProjectsMapContentSerializer(serializers.ModelSerializer):
         projects = Project.objects.filter(
             campaign_ended__gte=self.context['start_date'],
             campaign_ended__lte=self.context['end_date'],
-            status__slug__in=['done-complete', 'done-incomplete'])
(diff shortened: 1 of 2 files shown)


bluebottle/cms/serializers.py
```diff
@@ -130,7 +130,10 @@ def get_projects(self, obj):
         projects = Project.objects.filter(
             campaign_ended__gte=self.context['start_date'],
             campaign_ended__lte=self.context['end_date'],
-            status__slug__in=['done-complete', 'done-incomplete'])
+            status__slug__in=['done-complete', 'done-incomplete']
+        ).order_by(
+            'campaign_ended'
```
ashutoshb: @eodolphi : If we have to have the most recent projects at the top (descending order), then shouldn't then this be '-campaign_ended' ? [django docs](https://docs.djangoproject.com/en/1.10/ref/models/querysets/#order-by)

eodolphi: The frontend expects the projects to be ascending.

PR #2602, opened 2017-03-07
fix importing of record by date updated

ashutoshb:


eodolphi:
This looks like it should need a test that would have failed with the old code. 

But I guess this needs to go in quickly, so merging anyway.

PR #2606, opened 2017-03-08
hotfix/fix timestamp for vote influxdb log entry

ashutoshb:


bluebottle/analytics/tests/test_models.py
```diff
@@ -57,7 +57,7 @@ def setUp(self):
         }
 
     def test_country_tag(self, queue_mock):
-        ProjectFactory.create(theme=self.theme, status=self.status,
+        project = ProjectFactory.create(theme=self.theme, status=self.status,
                               country=self.country, location=None)
```
gannetson: This var isn't used, right? Otherwise indentation is off.

ashutoshb: 👍

bluebottle/analytics/management/commands/export_analytics_data.py
```diff
@@ -62,17 +62,23 @@ def handle(self, **options):
     @override_settings(ANALYTICS_ENABLED=True)
     def _process(self, cls, start_date, end_date):
         cls_name = cls.__name__
+        # TODO: Come up with a better way of generically getting the timestamp for influxdb log entry
```
gannetson: We might event want to change the way we set timestamps on different models. Having consistent naming and logic on when, what timestamp is set. For now this will do.

ashutoshb: Yes that would be very much preferred to have a consistent way of setting created and updated fields across all models.

PR #2607, opened 2017-03-08
add extra fields to project phase log and task member status log

ashutoshb:


bluebottle/tasks/models.py
```diff
@@ -371,6 +371,10 @@ class Analytics:
             'task_id': 'task_member.task.id'
         }
 
+        def extra_fields(self, obj, created):
+            # Force the time_spent to an int.
+            return {'hours': int(obj.task_member.time_spent)}
```
gannetson: I think ```time_spent could``` also be fractions of an hour. Maybe not used that often atm though.

ashutoshb: I just carried it over from how it was done for TaskMember analytics. time_spent is a PositiveSmallIntegerField.

gannetson: Cool. We will fix that at some other time then :-)

PR #2627, opened 2017-03-15
When a payout fails to create due to validation errors, show a message

eodolphi:
to the user about that validation error.

BB-9139 #resolve

### DIFF
--- a/bluebottle/payouts_dorado/adapters.py
+++ b/bluebottle/payouts_dorado/adapters.py
@@ -1,3 +1,4 @@
+import json
 import requests
 
 from django.core.exceptions import ImproperlyConfigured
@@ -5,6 +6,14 @@ from django.db import connection
 from requests.exceptions import MissingSchema
 
 from bluebottle.clients import properties
+
+
+class PayoutValidationError(Exception):
+    pass
+
+
+class PayoutCreationError(Exception):
+    pass
 
 
 class DoradoPayoutAdapter(object):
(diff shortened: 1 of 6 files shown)


bluebottle/projects/models.py
```diff
@@ -718,15 +716,6 @@ def project_post_init(sender, instance, **kwargs):
...
           dispatch_uid="bluebottle.projects.Project.post_save")
 def project_post_save(sender, instance, **kwargs):
-
-    if instance.payout_status == 'approved':
-        adapter = DoradoPayoutAdapter(instance)
-        try:
-            adapter.trigger_payout()
-        except ImproperlyConfigured:
-            logger.warning('Dorado not configured when project payout approved',
-                           exc_info=1)
```
gannetson: Why is this logger removed? Maybe add it where you catch this error in projects/admin ?

bluebottle/projects/admin.py
```diff
@@ -252,12 +254,27 @@ def get_urls(self):
...
+        if not request.user.has_perm('projects.approve_payout'):
+            return HttpResponseForbidden('Missing permission: projects.approve_payout')
+        elif project.payout_status != 'needs_approval':
+            self.message_user(request, 'The payout does not have the status "needs approval"')
+        else:
+            adapter = DoradoPayoutAdapter(project)
+            try:
+                adapter.trigger_payout()
+            except ImproperlyConfigured:
+                pass
```
gannetson: Be clear about errors creating a payout (also see below)

bluebottle/payouts_dorado/adapters.py
```diff
@@ -23,11 +32,15 @@ def trigger_payout(self):
...
-            #
+            response.raise_for_status()
+
             self.project.payout_status = 'created'
             self.project.save()
+        except requests.HTTPError:
+            try:
+                raise PayoutValidationError(json.loads(response.content))
+            except ValueError:
+                raise PayoutCreationError(response.content)
```
ashutoshb: If there is an payout creation error, we could reset the payout_status to its previous state.

gannetson:
Now we have merge errors. Probably because I merged @ashutoshb 's pr :-s

eodolphi:
I merged with the integration branch. Should be good now.

PR #2628, opened 2017-03-15
hide approve payout button if the campaign has not ended

ashutoshb:


gannetson:
Not sure if this is the best way to fix the 'problem'. Maybe it's better to change the logic where we set `payout_status` to `needs_approval` and only add that status if the project is finished and has a timestamp for `campaign_ended`. Or we should investigate how a project can have a final status without a `campaign_ended` date.

ashutoshb:
@gannetson : [views.py](https://github.com/onepercentclub/bluebottle/blob/feature/hide-approve-payout-button/bluebottle/payouts_dorado/views.py#L12) states that dorado can only ask for projects which have campaign ended set. If the campaign ended is not set, it returns a 404. My understanding was that the payouts can only happen once the campaign had ended and hence the approve payout button should be shown in the admin if and only if the campaign has ended.

gannetson:
Dorado will only try to get projects from Bb when a signal is send by Bb (after payout_status changes from `needs_approval` to `approved`). Bb should only send that signal when a project is ready for payout e.g. right status and timestamps. So I still think we should only set `payout_status` to `needs_approval` if all those requirements are met.
What do you think @ashutoshb ?

ashutoshb:
@gannetson : makes perfect sense. In order to setup my test environment, i was running the ```update projects_project set payout_status = 'needs_approval';``` which meant that the approve payout button showed up on all project details page.

PR #2633, opened 2017-03-17
Add retry when payout request fails

rollick:
BB-9312 #resolve


eodolphi:
See the discussion between me and loek in slack this morning.

It is probably not a good idea to merge this.

PR #2639, opened 2017-03-21
Feature/zaad

gannetson:


bluebottle/payments_telesom/tests/test_adapters.py
```diff
@@ -0,0 +1,70 @@
...
+
+from django.test.utils import override_settings
+
+from bluebottle.payments_telesom.adapters import TelesomPaymentAdapter
+from bluebottle.test.factory_models.donations import DonationFactory
+from bluebottle.test.factory_models.orders import OrderFactory
+from bluebottle.test.factory_models.payments import OrderPaymentFactory
+from bluebottle.test.utils import BluebottleTestCase
+
+flutterwave_settings = {
```
eodolphi: This should probably be call zaad_settings.

setup.py
```diff
@@ -74,6 +74,7 @@ def read_file(name):
     'gunicorn==19.5.0',
     'html5lib==0.9999999',
     'influxdb==3.0.0',
+    'ipgetter==0.6',
```
eodolphi: I do not see this imported anywhere, or am I missing something?

gannetson: Yup. Telesom is asking for an IP in their API, so I implemented a remote IP getter, but then it turned out you should always return ":1"... :-o

PR #2656, opened 2017-03-28
Refactor organizations api to remove bb_ code

rollick:
BB-9402 #resolve


gannetson:
Maybe we temporarily add back-up urls for the old api endpoints? Just duplicates of `api/organizations`. Or will the current front-end break with this code any way? Would be nice if we could have it backwards compatible if it's easy enough.

PR #2668, opened 2017-04-05
fix result page statistics

ashutoshb:


bluebottle/statistics/tests/test_unit.py
```diff
@@ -103,7 +103,7 @@ def test_project_incomplete_stats(self):
             online=0,
             involved=1
         )
-        self.assertEqual(self.stats.projects_realized, 1)
+        self.assertEqual(self.stats.projects_realized, 0)
```
rollick: @ashutoshb It would be good to transition some projects to realized so we can test the metric. I guess this means just generating status logs, or mocking something but that might be a bit complicated.

PR #2680, opened 2017-04-19
Fix chargebacks not being included

eodolphi:


gannetson:
@eodolphi Should we write tests for this by now?

PR #2682, opened 2017-04-19
added a deadline_to_apply field to tasks

mauricekleine:


mauricekleine:
I just noticed that this is not yet added in the backoffice, if that is desired.

PR #2685, opened 2017-04-21
Make it possible to read organiations when not logged in

eodolphi:
BB-9676 #resolve


rollick:
@eodolphi I think we should embed an org preview record (name, website) in the project response rather than opening up the get endpoint to anyone.

eodolphi:
That is fine by me, but what is the reasoning behind that exactly? I will add a comment with that then.

PR #2693, opened 2017-05-02
Add mpesa flow for Flutterwave

gannetson:


bluebottle/payments/serializers.py
```diff
@@ -5,9 +5,11 @@
 
 
 class OrderPaymentActionSerializer(serializers.ModelSerializer):
+    data = serializers.JSONField(source='payload')
+
     class Meta:
         model = OrderPaymentAction
-        fields = ('type', 'method', 'url', 'payload')
+        fields = ('type', 'method', 'url', 'payload', 'data')
```
gannetson: I want to sent a proper JSON object here `data`, but want to keep it backwards compatible so also keep the `payload` object.

bluebottle/payments_flutterwave/adapters.py
```diff
@@ -18,15 +20,44 @@
 
 
 class FlutterwavePaymentAdapter(BasePaymentAdapter):
-    MODEL_CLASSES = [FlutterwavePayment]
+    MODEL_CLASSES = [FlutterwavePayment, FlutterwaveMpesaPayment]
```
eodolphi: Wouldn't it be a little simpler if we had one model per adapter, and have both the MPesa en the Verve implementation of flutterwave extend from a common base adapter?

gannetson: We now we use the psp name to determine what adapter to use. We do the same with Docdata CC/Ideal vs Docdata Directdebit. I totally agree that this would be more readable if we would have separate adapters. Should we then split it and have adapters called DocdataDirectdebit, DocdataCreditcard and so on?

PR #2695, opened 2017-05-02
update cron job start and end date to a single day

ashutoshb:


bluebottle/analytics/tasks.py
```diff
@@ -42,8 +42,7 @@ def _process_handler(backend, timestamp, tags, fields):
 @shared_task
 def generate_engagement_metrics():
     today = date.today()
```
rollick: Maybe this should be datetime.datetime.utcnow().date() otherwise it will use local time. In the past we changed the timezone from UTC to Amsterdam which would mess with things. Not likely to happen often but still good to think about.

PR #2705, opened 2017-05-05
update the engagement metrics every 30 mins

ashutoshb:


bluebottle/analytics/tasks.py
```diff
@@ -41,8 +41,13 @@ def _process_handler(backend, timestamp, tags, fields):
 
 @shared_task
 def generate_engagement_metrics():
+    # Generate metrics every 30 mins. The timestamp will be of the end date and hence in the future.
```
rollick: @ashutoshb is this description correct? Isn't this generating metrics for the day so far?

ashutoshb: The cronjob will be run every 30 mins with the todays date as start date and tomorrows day as the end date. Hence with every run, we will update the current engagement number. The timestamp of the influx entry is always the end date. So with every run the current engagement number will just be over written.

PR #2729, opened 2017-05-12
Send an email to the taskmembers when the deadline of the project

eodolphi:
changed.

BB-9718 #resolve

### DIFF
--- a/bluebottle/tasks/taskmail.py
+++ b/bluebottle/tasks/taskmail.py
@@ -2,7 +2,7 @@ from celery import shared_task
 
 from django.dispatch import receiver
 from django.db import connection
-from django.db.models.signals import post_save, pre_delete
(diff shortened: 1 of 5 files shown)


bluebottle/tasks/taskmail.py
```diff
@@ -212,6 +212,30 @@ def new_reaction_notification(sender, instance, created, **kwargs):
...
+                    subject=subject,
+                    title=instance.title,
+                    original_date=previous_instance.deadline,
+                    date=instance.deadline,
+                    to=task_member.member,
+                    site=tenant_url(),
+                    link='/go/tasks/{0}'.format(instance.id)
+                )
+
+
```
gannetson: Hm, I would say that the signal should go in `signals.py`, but since we already did like this it's fine for now...

PR #2772, opened 2017-06-02
Make sure old project and tasks show in statics

gannetson:


ashutoshb:
@gannetson : I haven't checked if the old task members are missing in the task member status log. If so, then should we also consider doing the migration for the missing task member -> task member status log in this PR itself or maybe make a new PR for it. The rest of the PR looks good.

gannetson:
I haven't looked at TaskMembers, makes sense to do a data migration for that too. I'll look into it. 
So are you confident this migration won't mess up metrics in the sense that it will trigger events? @ashutoshb

ashutoshb:
@gannetson : there are no influxdb entries sent on trigger events on *logs Models. 👍

gannetson:
@ashutoshb, added task member log too

PR #2775, opened 2017-06-09
Hotfix/booking analytics dump

ashutoshb:
@rollick / @eodolphi : This special booking participation report has to be first passed by Frank thru Diwy for their approval and then thru Sebastian to confirm if this is exactly what they wish for. There might be  going to be a lot of ping-pong on this one, so I'll keep updating it as the change requests come in. Hence also I am not putting effort in writing any tests or code refactoring, till we nail the final specifics down.


bluebottle/statistics/tests/test_unit.py
```diff
@@ -398,3 +399,133 @@ def test_since_last_year(self):
...
+        count = self.statistics.get_tasks_status_count_by_location_group(location_group=self.location.group.name,
+                                                                         statuses=['in progress'])
+        self.assertEqual(count, 1)
+
+    def test_tasks_status_count_by_theme(self):
+        count = self.statistics.get_tasks_status_count_by_theme(theme='education', statuses=['in progress'])
+        self.assertEqual(count, 2)
+
+    def test_task_members_total(self):
+        pass
```
eodolphi: these should probably be filled in.

PR #2792, opened 2017-07-05
BB-10083 Replace Bank BIC field with generic bank details 

ashutoshb:


bluebottle/bb_projects/models.py
```diff
@@ -236,6 +235,10 @@ def people_needed(self):
 
         return requested - counts['total'] + (counts['externals'] or 0)
 
+    @property
+    def account_bic(self):
+        return self.account_details
+
```
gannetson: I guess this is here because of backwards compatibility, but this isn't writeable. Do you think we still have the desired behaviour? @ashutoshb

gannetson: Never mind. I see you have a test proving that it works :-)

gannetson: Jeez-Louis, I should not do reviews after such a long day. Your failing test prove I might be right! 
I created a PR against this which fixes it.
Nice work @ashutoshb. I like how you thought about keeping this compatible with current Reef!

PR #2797, opened 2017-07-10
Clean up and add tests for bulk importer BB-10100

rollick:


bluebottle/clients/management/commands/bulk_import.py
```diff
@@ -31,229 +31,195 @@
...
+    def reset(self, total=1):
+        self.total = total
+        self.count = 1
+
+
 class Command(BaseCommand):
+    """
+    TODO: json import should:
+        * pages:
+        ** always include language
```
gannetson: Or default to the tenant default?

PR #2798, opened 2017-07-14
BB-10086 extend category admin with cards

ashutoshb:


bluebottle/categories/models.py
```diff
@@ -43,3 +36,20 @@ def __unicode__(self):
...
+
+
+class CategoryContent(models.Model):
+    category = models.ForeignKey(Category, related_name='contents')
+    title = models.CharField(_('title'), max_length=60)
+    description = models.TextField(_("description"), max_length=190)
+    image = ImageField(_("image"), max_length=255, blank=True, null=True, upload_to='categories/content/',
+                       help_text=_("Category content image"))
+    video_url = models.URLField(max_length=100, blank=True, default='')
+    link = models.URLField(_('Link url'), blank=True)
```
gannetson: I think you also need a title to the link, so you can say "View all projects" or "Read more". Maybe check with Rolf?

ashutoshb:
As discussed with Rolf, the following changes need to be made.

- Add extra link_text field with 'Read more' as default
- Description is optional
- Show a message under the video url stating that 'The image is over-ridden if video url is present'

bluebottle/categories/admin.py
```diff
@@ -1,13 +1,29 @@
...
 from sorl.thumbnail.admin import AdminImageMixin
+from tinymce.widgets import TinyMCE
+from .models import Category, CategoryContent
+
+
+class CatergoryContentForm(forms.ModelForm):
+    description = forms.CharField(widget=TinyMCE())
+
+    class Meta:
+        mode = CategoryContent
```
eodolphi: This should probably be model right?

PR #2800, opened 2017-07-14
Make sure rewards get added to donations

gannetson:


bluebottle/clients/management/commands/bulk_import.py
```diff
@@ -54,18 +53,35 @@ def add_arguments(self, parser):
                             help="JSON import file.")
         parser.add_argument('--tenant', '-t', action='store', dest='tenant',
                             help="The tenant to run the recurring donations for.")
+        parser.add_argument('--models', '-m', action='store', dest='models',
```
ashutoshb: I am not sure of the lifetime of this bulk importer. If we envision keeping it around even after the vjb migration, then it would be nice to validate the inputs via the command line. see here: [e.g.](https://github.com/onepercentclub/bluebottle/blob/master/bluebottle/analytics/management/commands/export_engagement_metrics.py#L59)

bluebottle/clients/management/commands/bulk_import.py
```diff
@@ -75,8 +91,7 @@ def handle(self, *args, **options):
                         method_to_call(value)
                     print(" Done!\n")
 
-    def _generic_import(self, instance, data, excludes=False):
-        excludes = excludes or []
+    def _generic_import(self, instance, data, excludes=[]):
```
ashutoshb: its advisable to not use mutables as default params for functions to avoid unforseen side-effects.
[python gotchas](http://python-guide-pt-br.readthedocs.io/en/latest/writing/gotchas/)

bluebottle/clients/management/commands/bulk_import.py
```diff
@@ -252,10 +270,18 @@ def _handle_orders(self, data):
         if data['donations']:
             for don in data['donations']:
                 try:
-                    project = Project.objects.get(title=don['project'])
+                    project = Project.objects.get(slug=don['project'])
                 except Project.DoesNotExist:
-                    project = Project.objects.all()[0]
+                    print "Could not find project {0}".format(don['project'])
```
ashutoshb: useless nitpicking really, but it would be nice to use either log statements or use print consistently ala line 53

gannetson: I switched to prints to logging. Later we can decide where/how to show them.

PR #2815, opened 2017-07-27
fix export data

ashutoshb:
- Fixes time spent in export data
- FIxes date status change in export data
- Generic refactoring


bluebottle/settings/base.py
```diff
@@ -718,18 +718,6 @@
...
-        #         'id',
-        #         ('get_status_display', 'Status'),
-        #         'title',
-        #         'org_email',
-        #         'project__location',
-        #         'created',
-        #         'updated',
-        #     ),
-        #     'title': 'Suggestions',
-        # })
```
gannetson: Yeah! If you don't use it lose it.

bluebottle/tasks/models.py
```diff
@@ -143,6 +131,43 @@ def people_accepted(self):
             total_externals += member.externals
         return members.count() + total_externals
 
+    @property
+    def date_realized(self):
```
gannetson: Maybe a small comment here about how we calculate this?

PR #2817, opened 2017-07-28
category blocks sort order

ashutoshb:


gannetson:
Maybe we could use SortableMixin like we do for stats in results page to make the content blocks sortable?
https://github.com/onepercentclub/bluebottle/blob/master/bluebottle/cms/admin.py#L15

ashutoshb:
@gannetson : https://github.com/onepercentclub/bluebottle/pull/2818

PR #2818, opened 2017-07-31
make content blocks sortable in backoffice

ashutoshb:


ashutoshb:
@gannetson : Thanks for the headsup. Fixed the failing tests.

PR #2822, opened 2017-08-01
Extra permissions work

rollick:


bluebottle/projects/models.py
```diff
@@ -705,6 +708,14 @@ class Meta(BaseProject.Meta):
             ('api_add_project', 'Can add projects through the API'),
             ('api_change_project', 'Can change projects through the API'),
             ('api_delete_project', 'Can delete projects through the API'),
+            ('api_read_projectdocument', 'Can view project documents through the API'),
+            ('api_add_projectdocument', 'Can add project documents through the API'),
```
rollick: @gannetson not so sure about this but the project document ownership is based on the project ownership so the permissions needed to be here. It's a WIP

gannetson: Maybe we can't the default parent-permission model here because `ProjectDocument` should not be readable by anyone other than initiator (and perhaps reviewer). Not sure how it is solved now.

bluebottle/projects/tests/test_api.py
```diff
@@ -607,6 +607,7 @@ def test_project_create(self):
         project_data['slug'] = 'a-new-slug-should-not-be-possible'
         response_2 = self.client.put(project_url, project_data,
                                      token=self.another_user_token)
+        print(response_2.content)
```
gannetson: No!

PR #2837, opened 2017-08-08
Master into integration

rollick:


bluebottle/bb_tasks/permissions.py
```diff
@@ -93,3 +96,34 @@ def has_object_permission(self, request, view, obj):
             request.user == obj.task.author or
             request.user.is_staff
         )
+
+
+class IsRelatedToActiveProjectOrReadOnly(permissions.BasePermission):
```
gannetson: Don't we want to use `utils.BasePermission` for this?

PR #2849, opened 2017-08-14
Permissions refactor...

rollick:


bluebottle/bb_tasks/permissions.py
```diff
@@ -21,40 +21,48 @@ def get_parent_from_request(self, request):
...
-    def has_object_method_permission(self, method, user, view, obj):
+    def has_object_action_permission(self, action, user, obj):
         return user == obj.owner
 
 
-class MemberOrTaskOwnerOrReadOnlyPermission(BasePermission):
-    # TODO: Move this check to the serialiser
-    def _time_spent_updated(self, request, task_member):
-        if request.data:
+class TaskMemberTimeSpentPermission(BasePermission):
```
rollick: @eodolphi this permissions class won't work with the PermissionField because the logic is located in the has_object_permission and not in has_object_action_permission. This is like a field level permission - something to think about later.

bluebottle/utils/serializers.py
```diff
@@ -221,33 +221,15 @@ class PermissionField(BasePermissionField):
...
-        for key, attr in self.data_mappings.iteritems():
-            data[key] = getattr(value, attr)
-
-        request.data = data
-        view.request = request
-
-        return all(perm.has_method_permission(
-            method, user, view
+        return all(perm.has_action_permission(
+            method, user, view.model, value
```
rollick: The `value` here is assumed to be the parent...

bluebottle/wallposts/views.py
```diff
@@ -40,6 +41,10 @@ class WallpostList(ListAPIView):
     serializer_class = WallpostSerializer
     pagination_class = BluebottlePagination
 
+    # FIXME: figure out why it is necessary to include the TenantConditionalOpenClose
```
rollick: WIP...

bluebottle/bb_tasks/permissions.py
eodolphi: I do not think we want this to be handled by permissions right? This should be handled in the serializer.

PR #2858, opened 2017-08-17
Send more comple user profile as task owner

gannetson:
Now we send same profile as we already send for project owner


rollick:
@loek this should be against release/permissions

PR #2860, opened 2017-08-17
Filter my projects / tasks based on project role

rollick:
BB-10301


eodolphi:
I think we also filter the normal project list and task list resources. 

And possibly rewards / taskmembers / donations  etc.

PR #2879, opened 2017-08-28
Fix project initiator having edit right on tasks

eodolphi:


bluebottle/bb_tasks/tests/test_api.py
```diff
@@ -387,12 +385,6 @@ def test_task_member_set_time_spent(self):
                                     token=task_member_user_token)
         self.assertEqual(response2.status_code, status.HTTP_400_BAD_REQUEST)
 
-        # Project owner cannot update the time_spent
-        response3 = self.client.put('{0}{1}'.format(self.task_members_url, task_member.id),
-                                    {'time_spent': 5, 'task': task.id},
-                                    token=self.some_token)
-        self.assertEqual(response3.status_code, status.HTTP_403_FORBIDDEN, response3.data)
-
```
gannetson: Why remove this test? I understand that it will fail because now the project owner is the same user as the task owner (and that will be granted permission), but we should check that the project owner can't change it if it is another user.

eodolphi: I was pretty sure we check that in another test.
.
I will change it so that the project owner is not the task manager, and then it should fail. Duplicate tests are not really a bad thing.

PR #2883, opened 2017-08-29
Make sure we don't send the same task mail twice

gannetson:
BB-10426 #fixes


bluebottle/tasks/taskmail.py
```diff
@@ -194,16 +194,19 @@ def send_deadline_to_apply_passed_mail(task, subject, tenant):
...
-        send_mail(
-            template_name='tasks/mails/{}.mail'.format(template),
-            subject=subject,
-            task=task,
-            to=task.author,
-            site=tenant_url(),
-            edit_link='/tasks/{0}/edit'.format(task.id),
-            link='/tasks/{0}'.format(task.id),
+        if not task.mail_logs.filter(type=template).exists():
+            print template
```
eodolphi: This needs to go.

bluebottle/tasks/models.py
```diff
@@ -181,22 +186,20 @@ def deadline_to_apply_reached(self):
...
         if self.status == self.TaskStatuses.open:
             if self.people_applied:
                 if self.people_applied + self.externals_applied < self.people_needed:
                     self.people_needed = self.people_applied
-
                 if self.type == self.TaskTypes.ongoing:
-                    self.set_in_progress()
+                    self.status = self.TaskStatuses.in_progress
                 else:
-                    self.set_full()
```
eodolphi: This pattern is used one more time, also calling save directly and redundantly after the set. Maybe remove it everywhere?

bluebottle/tasks/taskmail.py
gannetson: It's nice and informative! We are open!

PR #2918, opened 2017-09-20
Add refunds

eodolphi:
Make it possible to refund docdata payments. Other payment methods will
fail.

Can be enabled by a tenant property (ENABLE_REFUNDS). User will need
'payments.refund_orderpayment' permissions. This is automatically
assigned to the "Finance" group in a migration.

Adds a button in the project and order payment admin to perform a
refund.

BB-9980 #resolve


bluebottle/settings/base.py
```diff
@@ -825,3 +825,5 @@
     'plus_one': False,
     'show_accepting': True
 }
+
+ENABLE_REFUNDS = True
```
gannetson: I guess we want to set this to False by default?

PR #2922, opened 2017-09-25
Fix donation permissions related to projects

eodolphi:
BB-10812 #resolve

### DIFF
--- a/bluebottle/projects/serializers.py
+++ b/bluebottle/projects/serializers.py
@@ -81,7 +81,7 @@ class ProjectPermissionsSerializer(serializers.Serializer):
         return obj
 
     rewards = RelatedResourcePermissionField('reward-list')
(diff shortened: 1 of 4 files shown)


bluebottle/projects/serializers.py
```diff
@@ -81,7 +81,7 @@ def get_attribute(self, obj):
         return obj
 
     rewards = RelatedResourcePermissionField('reward-list')
-    donations = ResourcePermissionField('order-manage-list')
```
rollick: Does this change need a test?

eodolphi: I guess it should.

I'll add one.

eodolphi: I added some tests that failed before the fix and passes after

PR #2927, opened 2017-09-26
Updates for user creation in the BO

rollick:
- send password reset email when creating user
- user is now active by default when created
- hide voting and address details when creating user
BB-10287 #resolve

### DIFF
--- a/bluebottle/bb_accounts/utils.py
+++ b/bluebottle/bb_accounts/utils.py
@@ -1,6 +1,8 @@
 import re
 
 from django.utils.translation import ugettext_lazy as _
+from django.utils.http import int_to_base36
+from django.contrib.auth.tokens import default_token_generator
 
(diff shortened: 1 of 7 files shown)


eodolphi:
What is the status of this? we should get this deployed somewhere this week right?

rollick:
@eodolphi It's waiting for QA to schedule it

bluebottle/bb_accounts/utils.py
```diff
@@ -20,15 +22,26 @@ def send_welcome_mail(user=None):
...
     }
 
-    send_mail(
-        template_name='bb_accounts/activation_email',
-        subject=subject,
-        to=user,
-        **c
-    )
+    # If there is no password set then use the welcome + password set template
+    if not user.password:
```
gannetson: What will happen if someone logs in with Facebook? Will they get the _set your password_ mail?

PR #2929, opened 2017-09-26
Feature/vjb payments

gannetson:


bluebottle/payments_external/tests/factory_models.py
```diff
@@ -0,0 +1,22 @@
+import factory
+from moneyed.classes import Money, USD
+
+from bluebottle.test.factory_models.payments import OrderPaymentFactory, OrderFactory
+
+from ..models import TelesomPayment
+
+
+class TelesomOrderFactory(OrderFactory):
+    total = Money(10, USD)
```
eodolphi: These should be called ExternalPayment....

bluebottle/clients/management/commands/bulk_import.py
```diff
@@ -166,7 +170,11 @@ def _handle_projects(self, data):
         except Project.DoesNotExist:
             project = Project(slug=data['slug'])
 
-        project.owner = Member.objects.get(email=data['user'])
+        try:
+            project.owner = Member.objects.get(email=data['user'])
+        except Member.DoesNotExist:
+            project.owner = Member.objects.all()[0]
```
eodolphi: What is the first user normally? Maybe we can make a special user for this purpose.

gannetson: Yeah, it's a bit lame like this, but it also weird to do a look up for 'info@voorjebuurt.nl' here. Maybe use admin@example.com?

PR #2976, opened 2017-10-30
Payouts for sourcing projects

gannetson:
Please make this a thorough review. I had to change some core mechanisms...


bluebottle/projects/models.py
```diff
@@ -340,6 +326,11 @@ def save(self, *args, **kwargs):
         if self.payout_status == 're_scheduled' and self.campaign_paid_out:
             self.campaign_paid_out = None
 
+        # BB-11037 If the project is re-opened, payout-status should be cleaned
```
rollick: @gannetson not sure ticket numbers should be in the code. I think they are better in the commit history

PR #2998, opened 2017-11-13
Add setting that specifies if anonymous rewards are allowed

eodolphi:
BB-11091 #resolve


bluebottle/projects/migrations/0051_auto_20171113_1637.py
```diff
@@ -0,0 +1,46 @@
...
+        migrations.AlterField(
+            model_name='project',
+            name='theme',
+            field=models.ForeignKey(blank=True, null=True, on_delete=django.db.models.deletion.SET_NULL, to='bb_projects.ProjectTheme'),
+        ),
+        migrations.AlterField(
+            model_name='projectdocument',
+            name='file',
+            field=bluebottle.utils.fields.PrivateFileField(max_length=110, upload_to=b'private/private/private/projects/documents'),
+        ),
```
gannetson: This is wrong of course. It should be fixed in https://github.com/onepercentclub/bluebottle/pull/3002

gannetson: Maybe just remove the AlterField migrations and only keep the AddField?

PR #3039, opened 2017-12-05
Feature/extra fields

gannetson:


bluebottle/projects/signals.py
```diff
@@ -63,3 +65,14 @@ def send_new_role_email(sender, instance, **kwargs):
...
+
+@receiver(post_save, weak=False, sender='projects.CustomProjectFieldSettings')
+def project_custom_fields(sender, instance, **kwargs):
+    """
+    Generate empty extra fields for all projects so exports etc work as expected.
+    """
+    if getattr(properties, 'CELERY_RESULT_BACKEND', None):
+        populate_extra_fields.delay(connection.tenant, instance)
+    else:
+        populate_extra_fields(connection.tenant, instance)
```
eodolphi: Isn't it cleaner to not save the extra fields, but just make sure the export understands missing fields?

gannetson: I was contemplating that too. You could just do a lookup when exporting of course, but I thought it would make sense to have a complete record for every project including the extra fields (even empty ones). It does now introduce some extra code (signals, tasks) that is not very clear maybe....

eodolphi: And we have loads of empty rows in our db that do nothing. If they are deleted somehow, things will fail unexpectedly.

gannetson: I will clear it up... Readable code is better :-)

gannetson: @eodolphi Cleared

PR #3043, opened 2017-12-08
Hotfix/bulk import improvements

gannetson:


eodolphi:
In general:

How will this work after the import? How will new employees receive the custom fields? Or will we do an import on a regular basis?

bluebottle/utils/admin.py
```diff
@@ -91,10 +94,34 @@ def export_as_csv(modeladmin, request, queryset):
...
+                for field in CustomMemberFieldSettings.objects.all():
+                    labels.append(field.name)
+            writer.writerow(row)
 
         for obj in queryset:
-            writer.writerow([prep_field(request, obj, field, manyToManySep) for field in field_names])
+            row = [prep_field(request, obj, field, manyToManySep) for field in field_names]
+            # Write extra field data
+            if queryset.model is Project:
+                for field in CustomProjectFieldSettings.objects.all():
```
eodolphi: Tying a general export function to project specific models seems like a bad idea. Any chance we can clean this up a bit?

gannetson:
Employees can't update these fields themselves, for now. The platform manager can change this in admin. I don't think we want to do periodic updates/imports at this moment. If we want to have something like this we should do so through webhooks / API or such.

bluebottle/utils/admin.py
gannetson: True

PR #3054, opened 2017-12-21
Make it possible to extend the deadline of a task until after the

gannetson:
project deadline

When a task deadline is extend to after the project deadline, also move
the project deadline.

Project deadlines cannot be more then a year after the project started.

BB-11282 #resolve

### DIFF
--- a/bluebottle/tasks/serializers.py
+++ b/bluebottle/tasks/serializers.py
@@ -1,3 +1,4 @@
+from datetime import timedelta
 from rest_framework import serializers
 from django.utils.translation import ugettext_lazy as _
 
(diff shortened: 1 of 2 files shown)


bluebottle/tasks/serializers.py
```diff
@@ -146,6 +150,24 @@ class Meta:
             'type',
         )
 
+    def _check_project_deadline(self, instance, validated_data):
+        project = validated_data['project']
+        if instance.deadline > project.deadline:
+            project.deadline = instance.deadline
+            project.save()
```
gannetson: I think this should only update if it's a sourcing project and not for funding (or combined) projects.

PR #3084, opened 2018-01-17
Make sure there is no private information in anonymous donation emails.

eodolphi:
BB-10705 #resolve


bluebottle/donations/templates/donations/mails/new_oneoff_donation.mail.txt
```diff
@@ -3,12 +3,12 @@
 
 {% block content %}
     {% if pledged %}
-        {% blocktrans with donor_name=donation.user.first_name first_name=donation.project.owner.full_name slug=donation.project.slug %}
+        {% blocktrans with first_name=donation.project.owner.full_name slug=donation.project.slug %}
 
         Hello {{ first_name }}, you've just received an invoiced donation by {{ donor_name }}. The amount will be transfered directly to your bank account by {{donor_name}}. If not, please contact the user via {{ admin_email }}.
         {% endblocktrans %}
```
gannetson: It appears that pledge donation now have mails without donor_name. The text will have strage blanks. Or do I miss something?

gannetson: @eodolphi {{ donor_name }} should be re-added I think

eodolphi: No that was precisely the issue. In bluebottle.donations.donationmails we correctly set donor_name in the context. It will be "an anonymous person" if the donation is anonymous. However in the blocktrans blocks we would override that again, and set it to the user's name. 

By not adding it to the blocktrans definition, the donor_name will be correct and translated.

PR #3112, opened 2018-02-09
Move donation settings to back office

gannetson:
This will move `show_donation_amount` and `default_donation_amounts` to back office.
We should still adapt front-end to use new entry in config api: `platform.donations`


danielkroon:
@gannetson I added the 'Needs more work' label because there are 2 minor things.

First thing: Please fix the 'amountss' typo in the backoffice.
screen shot 2018-10-09 at 17 45 53

The second thing: the api sends to default amounts in to the frontend. Shouldn't we remove the old one:
screen shot 2018-10-09 at 17 39 21

@gannetson apart from that this feature works fine. Is there already a frontend ticket so frontend can do there part?

PR #3137, opened 2018-03-12
Add platform settings for lock down

gannetson:


bluebottle/auth/middleware.py
```diff
@@ -189,93 +189,68 @@ def process_request(self, request):
...
-        return response
+            page_data = {}
+            if not hasattr(form, 'show_form') or form.show_form():
+                page_data['form'] = form
+
+            response = render_to_response('lockdown/form.html', page_data)
+            response.status_code = 401
+            return response
+
+        return None
```
eodolphi: Couldn't we greatly simplify this? No need to look at the session or the submitted form. We only need to check for the header being present right?

PR #3165, opened 2018-03-22
Add permissions to jet menu

gannetson:


eodolphi:
Nice. I need exactly this too.

PR #3190, opened 2018-04-04
Make sure we take amount_extra into account when calculating

eodolphi:
amount_needed in the admin

BB-12218 #resolve


eodolphi:
Note: It might be better to do the calculation in the project model, so that the project also closes automatically. But I am not sure if that is desirable.

PR #3194, opened 2018-04-09
Task count should not filter on task status. We show the tasks tab to

eodolphi:
users with rejected tasks as well.


bluebottle/bb_accounts/models.py
```diff
@@ -292,8 +292,7 @@ def task_count(self):
         task member in
         """
         task_count = Task.objects.filter(author=self).count()
-        taskmember_count = TaskMember.objects.filter(
-            member=self, status__in=['applied', 'accepted', 'realized']).count()
```
gannetson: With this `task_count` will also count rejected & withdrawn?

eodolphi: Yes. It should count all task members.

PR #3232, opened 2018-04-25
Add fiscal year and language as user attributes in looker views. Rena…

eodolphi:
…me model


gannetson:
This needs to be recreated against `master` or another release branch. Other than that, good to go.

PR #3269, opened 2018-05-23
Set the tenant properties when settings the tenant on the connection.

eodolphi:
This makes sure we always have the correct properties set.


eodolphi:
Note:

We need to be careful when deploying this. The database engine setting needs to be changed. But only when we release. Not before, not after.

Maybe we can do a release that only adds the database engine (without the overriden set_tenant method), and then change the settings. And then release again after that is done with all the other changes.

PR #3278, opened 2018-05-25
Feature/dispose project documents simple

eodolphi:


bluebottle/projects/admin.py
```diff
@@ -243,7 +243,9 @@ def __init__(self, *args, **kwargs):
 
     def clean(self):
         if (
-            self.cleaned_data['status'].slug == 'campaign' and
+            self.cleaned_data['status'].slug not in (
+                'plan-new', 'plan-submitted', 'plan-needs-work'
```
gannetson: Maybe also allow the project to be _closed_ without bank details. Or turn it around and only check for 'cleaned_data['status'].slug == 'campaign'`?

PR #3323, opened 2018-06-11
Send email password reset link with front-end url

gannetson:
BB-1682 #resolve

### DIFF
--- a/bluebottle/members/admin.py
+++ b/bluebottle/members/admin.py
@@ -330,12 +330,6 @@ class MemberAdmin(UserAdmin):
         reset_mail_url = reverse('admin:auth_user_password_reset_mail', kwargs={'user_id': obj.id})
         properties.set_tenant(connection.tenant)
 
-        # Don't show frontend password reset mail link for closed sites.
-        if properties.CLOSED_SITE:
-            return format_html(
-                "<a href='{}'>{}</a>",
-                reset_form_url, _("Reset password form")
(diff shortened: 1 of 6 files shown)


bluebottle/members/admin.py
```diff
@@ -325,9 +328,19 @@ def following(self, obj):
     def reset_password(self, obj):
         reset_form_url = reverse('admin:auth_user_password_change', args=(obj.id, ))
         reset_mail_url = reverse('admin:auth_user_password_reset_mail', kwargs={'user_id': obj.id})
-        return format_html("<a href='{}'>{}</a>  | <a href='{}'>{}</a>  ",
-                           reset_form_url, _("Reset password form"),
-                           reset_mail_url, _("Send reset password mail"))
+        properties.set_tenant(connection.tenant)
+
+        # Don't show frontend password reset mail link for closed sites.
+        if properties.CLOSED_SITE:
```
eodolphi: This also means that partner users cannot reset passwords anymore. That does not seem desirable to me.

gannetson: I agree, but now it wasn't working for both.
My suggestion is to add another 'send admin pwd reset' mail link.

eodolphi: I do not think I understand. On closed platforms the reset password route is exposed, but just not applicable for most users, or am I missing something?

PR #3342, opened 2018-07-11
Hotfix/fix newsitem

gannetson:
Still trying to nail this and clean-up the different views and managers in used for filtering published content.


eodolphi:
This seems to miss migrations. Or did the models already have the same fields by accident?

eodolphi:
We probably need to test the language stuff for pages. It is a little unclear to me that that really works.

PR #3345, opened 2018-07-13
Add refund confirmation form

gannetson:


bluebottle/projects/admin.py
```diff
@@ -518,7 +516,8 @@ def approve_payout(self, request, pk=None):
 
         return HttpResponseRedirect(project_url)
 
-    def refund(self, request, pk=None):
+    @refund_confirmation_form(ProjectRefundForm)
```
eodolphi: Why is this a decorator? 

We only use it here, so this adds a useless level of indirection. 

If we have more places we want to use confirmation forms in the admin, this looks like a reasonable path, but then we need to make the decorator more generic.

PR #3361, opened 2018-08-02
Upgrade Flutterwave to v2 api

gannetson:


gannetson:
We could test this once Nigerian campaigns have ended. We also should have someone over there for testing.

PR #3381, opened 2018-08-28
Refunded donations are not cancelled. They should not be counted in the

eodolphi:
total amount in the frontend.

BB-12990 #resolve

### DIFF
--- a/bluebottle/projects/models.py
+++ b/bluebottle/projects/models.py
@@ -545,7 +545,6 @@ class Project(BaseProject, PreviousStatusMixin):
         return self.get_money_total([
             StatusDefinition.CANCELLED,
             StatusDefinition.REFUND_REQUESTED,
-            StatusDefinition.REFUNDED,
         ])
 
     @property


eodolphi:
@gannetson I do not really understand why you changed the definition of amount_cancelled, but it breaks how projects show in the frontend.

Maybe we should display another property in the backoffice? For now this fixes BB-12990

gannetson:
@eodolphi Because for (completely) refunded projects they want to show the initial amount raised. 
What would be the use of amount_cancelled?
My idea was dat the front-end would only show amount_cancelled and cancelled donations if the project status is set to 'refunded'.

eodolphi:
Amount cancelled is exactly that. The amount that is returned to the supporters after the project failed. For these projects is seems logical to me that the amount_raised is 0, since everything is returned...

danielkroon:
@eodolphi @gannetson I tested this on testing and it worked. Since the resolve date is today you can continue your discussing but we need this fix only today :) I discussed this with Ernst.

PR #3383, opened 2018-08-28
Account holder name cannot be longer then 100 characters.

eodolphi:
BB-12450 #resolve


gannetson:
Will this send user friendly error messages if we enter longer details on the front-end?

danielkroon:
@eodolphi works and merged!

PR #3384, opened 2018-08-28
Show "Anonymous user" instead of an empty name in donation export view

eodolphi:
BB-12529 #resolve


eodolphi:
I removed the label again, since that was not what this ticket is about. 

This is about the supporters export in the frontend. And I am pretty sure that is fixed with this pull request.

danielkroon:
@eodolphi ah my bad. I thought it was the backoffice export. It works, so it's merged.

PR #3386, opened 2018-08-29
Add all reviewers to project filter

gannetson:
@danielkroon If you can test this it can we released as a small fix.

Relates to https://onepercentclub.atlassian.net/browse/BB-13088


danielkroon:
@gannetson tested and merged!

PR #3394, opened 2018-09-06
Multiple fixes

gannetson:
* Move some donation settings to back-office (not enabled in front-end yet)
* Add all reviewers to reviewer filter in project admin: BB-13088
* Show _Anonymous user_ instead of empty in export: BB-12529
* Account holder name can't be more than 100 chars: BB-12450


danielkroon:
@gannetson I'm missing a message that I can only use 100 characters for https://onepercentclub.atlassian.net/browse/BB-12450. Now it's just impossible to add more characters.

danielkroon:
@gannetson why do we show multiple currency rows in the default amounts tab. If there is only one currency for this platform? See http://voor-je-buurt.localhost:8000/en/admin/donations/donationplatformsettings/1/change/#/tab/inline_0/
If we the platform has more currencies it makes sense. But now we have multiple default amount rows. Doesn't really make sense.
screen shot 2018-09-07 at 14 47 22

danielkroon:
@gannetson apart from the 2 comments above I didn't find anything strange.

eodolphi:
@danielkroon  I added a help text to the account holder field.

Your comment about the default currency is very hard to fix, and probably not worth the time we have to put in there. Nothing breaks, it is just a little weird. So I'd say: leave it like this.

PR #3401, opened 2018-09-18
Add geo location to task.

eodolphi:
Add new generic "Place" model that makes it possible to add a place to
a model. It includes address fields and geoposition.

TODO: also use this to store a project's location.

BB-13141 #resolve


bluebottle/tasks/models.py
```diff
@@ -101,6 +105,15 @@ def owner(self):
     def parent(self):
         return self.project
 
+    @property
+    def place(self):
+        try:
+            return self.places.get()
+        except Place.DoesNotExist:
+            return None
+        import ipdb; ipdb.set_trace()
```
gannetson: Say wot?

eodolphi: Haha, die kwam gewoon door de test natuurlijk.

eodolphi:
This still needs indexing of task places.

bluebottle/tasks/models.py
```diff
@@ -101,6 +105,13 @@ def owner(self):
     def parent(self):
         return self.project
 
+    @property
+    def place(self):
+        try:
+            return self.places.get()
```
gannetson: You could now have multiple places, right?
I think you should also catch `Place.MultipleObjectsReturned` or just do `return self.places.first()` (which is less explicit of course)

gannetson: Ah ne'er mind you limit it in admin too...

PR #3409, opened 2018-09-26
Online / on location filter should filter both null and blank fields.

eodolphi:


danielkroon:
@eodolphi tested and merged

PR #3411, opened 2018-10-02
Make the fundraiser image not required. This makes it possible to edit

eodolphi:
a fundraiser without sending an image.


eodolphi:
This makes the image field of fundraiser not required. That fixes updates of fundraiser without PUTting an image, but makes the image also not required during the creation of fundraisers. So this has to be deployed after the frontend fix that will make the field required during creation.

danielkroon:
@eodolphi I couldn't get this branch to work. Will try to test this monday.

PR #3414, opened 2018-10-03
Validate that tasks for projects with a duration are within the duration

eodolphi:
BB-13245 #resolve


danielkroon:
@eodolphi I can still choose task deadline that exceeds the project duration. Does frontend need to do something else before this fix works?

eodolphi:
No. How can you choose a deadline with out error? I cannot reproduce that...

PR #3415, opened 2018-10-15
Correctly activate tenant translations in API requests

eodolphi:
BB-12066 #resolve


gannetson:
Will we add a test for this later?

gannetson:
Will we add a test for this later?

PR #3419, opened 2018-10-24
Hotfix/token login

eodolphi:


eodolphi:
This also needs a frontend fix: https://github.com/onepercentclub/reef/pull/4687

PR #3422, opened 2018-10-26
Make sure that we catch exceptions when updating statuses in cron

eodolphi:
deadline reached.

Without this, we will not continue with the command if somethign fails, and other tenants
will not be updated.


eodolphi:
@gannetson I will not fix coverage for this pr. These lines a really hard to hit, and do not really matter. Can you have a look if the pull request makes any sense, and then please merge?

PR #3453, opened 2018-12-10
Remove migrations that do nothing

gannetson:


eodolphi:
This is already fixed in the original branch.

PR #3471, opened 2019-01-09
Fix permission check on plain payouts

eodolphi:


eodolphi:
This still needs tests. I can create them on friday, but if we want to release this quickly, @gannetson will need to add them tomorrow.

PR #3476, opened 2019-01-15
Use filetransfers for private files

gannetson:
Fixes https://onepercentclub.atlassian.net/browse/BB-13632

### DIFF
--- a/bluebottle/utils/views.py
+++ b/bluebottle/utils/views.py
@@ -1,16 +1,16 @@
+import mimetypes
 import os
 from collections import namedtuple
 
 from django.conf import settings
 from django.core.signing import TimestampSigner, BadSignature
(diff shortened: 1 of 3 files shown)


bluebottle/utils/views.py
```diff
@@ -244,13 +246,9 @@ def get_object(self):
 
     def get(self, request, *args, **kwargs):
         field = getattr(self.get_object(), self.field)
-        response = HttpResponse()
-        response['X-Accel-Redirect'] = field.url
```
gannetson: Not sure if removing this won't break file serving on testing/prod servers.
Do you know @eodolphi ?

gannetson: I re-added it, because tests were relying on it too, so it's probably of some good use ;-)

PR #3480, opened 2019-01-22
Feature/SCIM

eodolphi:


bluebottle/scim/tests/test_api.py
```diff
@@ -0,0 +1,667 @@
...
+
+
+class SCIMUserListTest(AuthenticatedSCIMEndpointTestCaseMixin, BluebottleTestCase):
+    @property
+    def url(self):
+        return reverse('scim-user-list')
+
+    def setUp(self):
+        for i in range(9):
+            BlueBottleUserFactory.create(is_superuser=False)
```
gannetson: Why not `BlueBottleUserFactory.create_batch(10, is_superuser=False)`?

PR #3491, opened 2019-02-05
Add project title and slug to stripe metadata.

eodolphi:
BB-13679 #resolve


danielkroon:
@eodolphi I created a bug for this pr https://onepercentclub.atlassian.net/browse/BB-13936

PR #3493, opened 2019-02-06
Set donation payout amount on create too

gannetson:


danielkroon:
@gannetson I tested this and it works. Can you fix the merge conflict and then merge it?

PR #3497, opened 2019-02-11
Do not store country of payout account, instead we use the value from

eodolphi:
the API.

BB-13893 #resolve
BB-13944 #resolve


eodolphi:
@danielkroon I am pretty sure this works as expected. But can you please verify that saving stripe kyc data still works? Also when you change the country half way through.

PR #3498, opened 2019-02-12
Hotfix/saml prevent provision filter users in group

eodolphi:


bluebottle/scim/tests/test_api.py
```diff
@@ -273,7 +273,7 @@ def url(self):
         return reverse('scim-user-list')
 
     def setUp(self):
-        for i in range(9):
+        for i in range(10):
             BlueBottleUserFactory.create(is_superuser=False)
```
gannetson: `BluebottleUserFactory.create_batch(10, ...)`

PR #3509, opened 2019-02-27
Set payment-descriptor for stripe payments

eodolphi:
BB-13861 #resolve


eodolphi:
@danielkroon Please test with a tenant with a long name. In stripe you should see the first 22 characters of  "<tenant name>-<project-id>" as the "statement descriptor.

gannetson:
@eodolphi Does this now apply for all payment methods?

PR #3533, opened 2019-03-18
Make sure that if no payout account is set we also trigger a validati…

eodolphi:
…on error. BB-14119 #resolve


danielkroon:
@eodolphi I still encounter this problem when checking out this pr https://www.useloom.com/share/eedcb14a3ba746e092b4cbe0910540c7

PR #3546, opened 2019-04-09
Make it possible to filter initiatives by owner

eodolphi:


eodolphi:
Besides a tiny feature in the first commit, this also fixes the tests for feature/initiatives

PR #3554, opened 2019-04-12
First go at image admin for initiatives

gannetson:


bluebottle/files/models.py
```diff
@@ -7,12 +7,6 @@
 
 class File(models.Model):
     id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
-    owner = models.ForeignKey(
```
eodolphi: I added this so that we can do permission checks. But since we are not doing that now, let's leave it like this.

bluebottle/initiatives/tests/test_api.py
```diff
@@ -144,6 +144,8 @@ def test_put_image(self):
             data['included'][0]['attributes']['links']['large'],
             HTTP_AUTHORIZATION="JWT {0}".format(self.owner.get_jwt_token())
         )
+        print data
```
eodolphi: the print statement should not be there.

bluebottle/files/serializers.py
```diff
@@ -5,21 +5,20 @@
...
 
 class FileSerializer(ModelSerializer):
     file = serializers.FileField(write_only=True)
     filename = serializers.SerializerMethodField()
-    size = serializers.IntegerField(read_only=True, source='file.size')
-    owner = ResourceRelatedField(read_only=True)
-
     links = serializers.SerializerMethodField()
+    owner = ResourceRelatedField(read_only=True)
+    size = serializers.IntegerField(read_only=True, source='file.size')
```
eodolphi: I think the links method for files should just return a link to the file. ImageSerializers should have overriden links method, that is the same as the method below.

gannetson:
> Can you team up with Jhon to make the nessecary changes to the frontend?

Done.

bluebottle/files/serializers.py
gannetson: Done

PR #3556, opened 2019-04-16
Feature/initiative messages

gannetson:


bluebottle/initiatives/admin.py
```diff
@@ -18,5 +19,8 @@ def get_fieldsets(self, request, obj=None):
             (_('Review'), {'fields': ('reviewer', 'review_status', 'review_status_transition')}),
         )
 
+    def get_inline_instances(self, request, obj=None):
+        return [MessageAdminInline(self.model, self.admin_site), ]
```
eodolphi: Why do we do this and don't we just set the inlines normally?

bluebottle/notifications/messages.py
```diff
@@ -0,0 +1,33 @@
...
+                recipient=recipient
+            ) for recipient in self.get_recipients()
+        ]
+
+    def get_recipients(self):
+        return [self.obj.owner]
+
+    def compose_and_send(self):
+        for message in self.get_messages():
+            message.send()
```
eodolphi: This should probably be reversed. First save and then send, and record that the sending took place. This way we can see if something went wrong from the database.

PR #3560, opened 2019-04-30
Make sure projects cannot go live with a deadline too far in to the

eodolphi:
future.

BB-14248 #resolve


danielkroon:
@eodolphi reopened see my comment in https://onepercentclub.atlassian.net/browse/BB-14248

PR #3571, opened 2019-05-13
Add some activity

gannetson:


bluebottle/activities/models.py
```diff
@@ -0,0 +1,98 @@
...
+
+    initiative = models.ForeignKey('initiatives.Initiative', related_name='activites')
+
+    title = models.CharField(_('title'), max_length=255, unique=True, db_index=True)
+    slug = models.SlugField(_('slug'), max_length=100, unique=True)
+    description = models.TextField(
+        _('description'), blank=True
+    )
+
+    image = ImageField(
```
eodolphi: I think that for now we do not need images / videos for activities.

bluebottle/activities/models.py
```diff
@@ -0,0 +1,98 @@
...
+        return self.contributions.count()
+
+    def save(self, **kwargs):
+        if not self.slug:
+            self.slug = slugify(self.title)
+
+        super(Activity, self).save(**kwargs)
+
+    @classmethod
+    def initiative_is_approved(cls, instance):
```
eodolphi: I know I probably wrote this :) but this looks a bit weird.

bluebottle/activities/serializers.py
```diff
@@ -0,0 +1,67 @@
...
+            'initiative'
+        ]
+        resource_name = 'activities'
+
+
+class ActivitySerializer(BaseActivitySerializer):
+    id = serializers.CharField(source='slug', read_only=True)
+    data = serializers.SerializerMethodField()
+
+    def get_data(self, instance):
```
eodolphi: This needs to go

bluebottle/activities/serializers.py
```diff
@@ -0,0 +1,67 @@
...
+
+    class JSONAPIMeta:
+        included_resources = [
+            'owner',
+            'initiative'
+        ]
+        resource_name = 'activities'
+
+
+class ActivitySerializer(BaseActivitySerializer):
```
eodolphi: I think we should use a polymorphic serializer here instead of this stuff. And I remember implementing this actually.

bluebottle/events/models.py
```diff
@@ -0,0 +1,196 @@
...
+from bluebottle.activities.models import Activity, Contribution
+from bluebottle.notifications.decorators import transition
+
+
+class Event(Activity):
+    start = models.DateTimeField(_('start'))
+    end = models.DateTimeField(_('end'))
+    registration_deadline = models.DateTimeField(_('registration deadline'))
+    capacity = models.PositiveIntegerField()
+    automatically_accept = models.BooleanField(default=True)
```
eodolphi: We should check with @Rolfertjan which fields we need, I  am pretty sure the `automatically accept` field should be there. And possibly the `registration_deadline` to.

bluebottle/events/views.py
```diff
@@ -0,0 +1,54 @@
...
+
+from bluebottle.activities.permissions import ActivityPermission
+from bluebottle.events.models import Event, Participant
+from bluebottle.events.serializers import EventSerializer, ParticipantSerializer
+from bluebottle.utils.permissions import (
+    OneOf, ResourcePermission, ResourceOwnerPermission)
+from bluebottle.utils.views import RetrieveUpdateAPIView, ListCreateAPIView, JsonApiViewMixin
+
+
+class EventList(JsonApiViewMixin, AutoPrefetchMixin, ListCreateAPIView):
```
eodolphi: I think we should use a Polymorphic activity  list endpoint instead of an endpoint specific for the activity type.

eodolphi: Possibly we do need an create endpoint for every activity type, because ember does not really like polymorphic endpoints that return a single resource.

PR #3575, opened 2019-05-20
Also send a mail to the object's owner when a MediaWallpost is created.

eodolphi:
Make sure we do not send the message when a donation is attached.

BB-14343 #resolve

### DIFF
--- a/bluebottle/wallposts/mails.py
+++ b/bluebottle/wallposts/mails.py
@@ -41,15 +41,16 @@ import logging
 from django.db.models.signals import post_save
 from django.dispatch import receiver
 
-from bluebottle.wallposts.models import Reaction, TextWallpost
+from bluebottle.wallposts.models import Reaction, TextWallpost, MediaWallpost
 from bluebottle.wallposts.notifiers import ObserversContainer
 
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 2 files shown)


bluebottle/wallposts/mails.py
```diff
@@ -41,15 +41,16 @@
...
 from bluebottle.wallposts.notifiers import ObserversContainer
 
 logger = logging.getLogger(__name__)
 
 
 @receiver(post_save, weak=False, sender=TextWallpost)
+@receiver(post_save, weak=False, sender=MediaWallpost)
 def new_wallpost_notification(sender, instance, created, **kwargs):
-    if created:
+    if created and not instance.donation:
```
gannetson: A `MediaWallpost` won't ever have a donation, right?

eodolphi: Apparently it does. If I do a donation and then post a message, a mediawallpost is created.

eodolphi: I do not think we ever make textwallposts anymore.

gannetson: I guess you want `sender=Wallpost` here. Now you won't get notified when someone writes a `TextWallpost`.
I think a donation wallpost is a `SystemWallpost` unless someone adds a message, then it turns into a `TextWallpost`.

PR #3595, opened 2019-06-03
Test participant transitions and side-effects

eodolphi:


bluebottle/events/models.py
```diff
@@ -149,6 +149,7 @@ def start(self, **kwargs):
     )
     def done(self):
         for member in self.participants:
+            member.activity = self
```
gannetson: This seems redundant

eodolphi: It is not completely redundant. The activity model of the member does not update when updating the activity. That will cause the condition on the member transition to fail.

I'll add a comment explaining this.

PR #3604, opened 2019-06-11
Do not update stripe payment after charge.

eodolphi:
Instead rely on the `charge.succeeded` event to update the payments.
This wat we do not create 2 wallposts and only send one email.

BB-14537 #resolve


eodolphi:
Note that this is hard to test locally since by default, webhook events do not reach your local server.

danielkroon:
@eodolphi I tried to test this locally but as you already said it's to hard. I merged it and will test this on staging.

Pieter-Rees:
@danielkroon Can you elaborate what's happening when you test this?

PR #3621, opened 2019-07-09
Add wallposts to initiatives and activities

gannetson:


eodolphi:
I guess this still needs including the wallposts in the initiatives / activity resource. But for now this is fine.

PR #3650, opened 2019-07-25
Only require location if 'is online' is False

gannetson:


danielkroon:
@gannetson please look at my comment in: https://onepercentclub.atlassian.net/browse/BB-14795

PR #3678, opened 2019-09-09
Improve money field

gannetson:


bluebottle/activities/transitions.py
```diff
@@ -16,14 +16,10 @@ def is_complete(self):
...
 
-    def is_activity_manager(self, user):
-        return self.instance.initiative.activity_manager == user
-
     @transition(
         source=ReviewTransitions.values.draft,
         target=ReviewTransitions.values.submitted,
-        conditions=[is_complete],
-        permissions=[is_activity_manager]
+        conditions=[is_complete]
```
eodolphi: Why would we remove this? Only managers should be able to do this.

gannetson: Shouldn't also others like initiative.owner be able to transition activities?

PR #3694, opened 2019-09-23
Feature/required and error fields

eodolphi:


bluebottle/funding_stripe/models.py
```diff
@@ -183,35 +184,80 @@ def payment_methods(self):
     DOCUMENT_SPEC = json.load(file)
 
 
+@memoize(timeout=60 * 60 * 24)
+def get_specs(country):
+    print '1111'
```
gannetson: No!

gannetson: Removed it. Awaiting tests.

PR #3734, opened 2019-10-07
Correctly return video url

eodolphi:


bluebottle/initiatives/serializers.py
```diff
@@ -88,9 +88,9 @@ class InitiativeSerializer(NoCommitMixin, ModelSerializer):
         ActivitySerializer, many=True, read_only=True
     )
     slug = serializers.CharField(read_only=True)
+    slug = serializers.CharField(required=False, allow_blank=True)
```
LiaVinci: I'm assuming the duplicated slug is by mistake

PR #3747, opened 2019-10-14
added specific prefix for ie

Pieter-Rees:


bluebottle/bluebottle_dashboard/static/jet/css/select2/_single.scss
```diff
@@ -27,6 +27,11 @@
       content: $icon-arrow-down;
       line-height: 32px;
     }
+
+    @media screen and (-ms-high-contrast: active),
+      screen and (-ms-high-contrast: none) {
+      top: 1rem;
```
Pieter-Rees: I saw 9 thousend and some more lines removed. I hope i compiled in a good way. It's all on one line now. It's uglyfied by default. It wasn't before.
Anyway, this is the fix

PR #3756, opened 2019-10-17
Use new donations for wallposts

gannetson:


bluebottle/fsm/__init__.py
```diff
@@ -172,6 +172,9 @@ def transition_to(self, transition, user=None, **kwargs):
                 transition=transition,
                 **kwargs
             )
+            if user:
+                from bluebottle.utils.admin import log_action
+                log_action(self.instance, user, 'Changed status to {} (API)'.format(transition.target))
```
LiaVinci: Are we suppose to leave logs in the code?

gannetson: In Django admin the actions are logged to a history so you can see who modified the record, how and when. So this is not like console.log ;-)

PR #3806, opened 2019-11-05
Limit status filter in activity/funding admin

gannetson:


scripts/generate_payouts.py
```diff
@@ -0,0 +1,31 @@
+# -*- coding: utf-8 -*-
```
eodolphi: This should not be here I think

PR #3882, opened 2019-11-19
Fix setting hours by task owner

gannetson:


bluebottle/assignments/views.py
```diff
@@ -82,6 +83,17 @@ class ApplicantDetail(JsonApiViewMixin, AutoPrefetchMixin, RetrieveUpdateAPIView
         'document': ['document'],
     }
 
+    def perform_update(self, serializer):
```
eodolphi: Don't we want to move this to the model, so that these transitions also happen in the back-office?

gannetson: Sounds good...

PR #3886, opened 2019-11-20
Try to submit all fundings when a payout account is submitted.

eodolphi:
Send message to owner when account is rejected / verified
Fix webhook, reject account when status becomes unverified.
Use status for verification check. This is much faster, since it does
not require a call to stripe.

BB-14634 #resolve


eodolphi:
We probably need to test this a little before deploying.

PR #3887, opened 2019-11-20
Try to submit all fundings when a payout account is submitted.

eodolphi:
Send message to owner when account is rejected / verified
Fix webhook, reject account when status becomes unverified.
Use status for verification check. This is much faster, since it does
not require a call to stripe.

BB-14634 #resolve

### DIFF
--- a/bluebottle/funding/models.py
+++ b/bluebottle/funding/models.py
@@ -572,6 +572,12 @@ class PayoutAccount(ValidatedModelMixin, PolymorphicModel, TransitionsMixin):
     updated = models.DateTimeField(auto_now=True)
     reviewed = models.BooleanField(default=False)
 
+    @property
+    def funding(self):
+        for account in self.external_accounts.all():
+            for funding in account.funding_set.all():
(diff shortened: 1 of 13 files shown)


eodolphi:
This probably needs some testing with actual webhooks before we release this.

bluebottle/funding/models.py
```diff
@@ -572,6 +572,12 @@ class PayoutAccount(ValidatedModelMixin, PolymorphicModel, TransitionsMixin):
     updated = models.DateTimeField(auto_now=True)
     reviewed = models.BooleanField(default=False)
 
+    @property
+    def funding(self):
+        for account in self.external_accounts.all():
+            for funding in account.funding_set.all():
+                return funding
```
gannetson: or `return account.external_accounts.first().funding_set.first()`?

eodolphi: That would work most of the time. But not if there would be accounts without fundings. In theory that could happen.

PR #3894, opened 2019-11-27
Add stats to the meta of an initiative detail representation.

eodolphi:
This will make it way easier to show stats on succeeded initiatives.

BB-15896 #resolve

### DIFF
--- a/bluebottle/activities/admin.py
+++ b/bluebottle/activities/admin.py
@@ -1,6 +1,7 @@
 from django.contrib import admin
 from django.contrib.admin import SimpleListFilter
 from django.db.models import Q
+from django.template import loader
 from django.urls import reverse
 from django.utils.html import format_html
 from django.utils.translation import ugettext_lazy as _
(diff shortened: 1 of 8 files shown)


bluebottle/initiatives/serializers.py
```diff
@@ -158,6 +166,21 @@ class JSONAPIMeta:
...
+
+        return {
+            'activities': len(activities),
+            'contributions': sum(stat['count'] for stat in stats),
+            'hours': sum(stat['hours'] or 0 for stat in stats if 'hours' in stat),
+            'amount': sum(
+                convert(Money(stat['amount']['amount'], stat['amount']['currency']), currency).amount
+                for stat in stats if 'amount' in stat
+            ),
+        }
```
gannetson: I guess we want to move the 'stats' calculation to the model if we want to show the complete stats in admin (or other locations) too. It's a bit unclear that we have different representations of 'stats' on model and serializer now.

bluebottle/activities/admin.py
```diff
@@ -117,10 +118,11 @@ def get_fieldsets(self, request, obj=None):
         )
 
     def stats_data(self, obj):
-        return format_html("<table>{}</table>", format_html("".join([
-            format_html(u"<tr><th>{}</th><td>{}</td></tr>", key, value) for key, value in obj.stats.items()
-        ])))
+        template = loader.get_template(
+            'admin/activity-stats.html'
+        )
```
gannetson: Yes, this is way nicer

bluebottle/assignments/models.py
```diff
@@ -181,7 +181,7 @@ def save(self, *args, **kwargs):
         elif self.status == ApplicantTransitions.values.failed and self.time_spent not in [None, '0', 0.0]:
             self.transitions.succeed()
         super(Applicant, self).save(*args, **kwargs)
-        if created:
+        if created and self.status == 'new':
```
gannetson: Let's stick with `ApplicantTransitions.values.new`

PR #3902, opened 2019-12-02
Do not index the location of activities that have `is_online=True`.

eodolphi:
This is easier then deleting the location in the save method, since we
do not want the location to disappear immediately.

BB-15764 #resolve

### DIFF
--- a/bluebottle/activities/documents.py
+++ b/bluebottle/activities/documents.py
@@ -114,11 +114,9 @@ class ActivityDocument(DocType):
         return len(instance.contributions.filter(status__in=('new', 'success')))
 
     def prepare_country(self, instance):
-        if hasattr(instance, 'location') and instance.location:
-            return instance.location.country_id
-        if hasattr(instance.initiative, 'location') and instance.initiative.location:
+        if instance.initiative.location:
             return instance.initiative.location.country_id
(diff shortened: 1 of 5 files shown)


bluebottle/activities/documents.py
```diff
@@ -114,11 +114,9 @@ def prepare_contribution_count(self, instance):
         return len(instance.contributions.filter(status__in=('new', 'success')))
 
     def prepare_country(self, instance):
-        if hasattr(instance, 'location') and instance.location:
-            return instance.location.country_id
```
gannetson: An activity might have a geolocation. Are you sure we only want to filter on initiative location?

bluebottle/activities/documents.py
```diff
@@ -133,9 +131,7 @@ def prepare_expertise(self, instance):
             return {'id': instance.expertise_id}
 
     def prepare_position(self, instance):
-        if hasattr(instance, 'location') and instance.location:
-            position = instance.location.position
-            return {'lat': position.get_y(), 'lon': position.get_x()}
```
gannetson: Same, an activity might have a location and you might want to filter on this.

bluebottle/activities/documents.py
gannetson: Ah you do so in specific activity :-) I see now

PR #3904, opened 2019-12-03
Add my controbutions end point.

eodolphi:
Only returns contributions of the current user. Returns all
contributions without pagination.


gannetson:
Are you sure we want to return **all** contributions? Also failed/incomplete donations and such?

PR #3914, opened 2019-12-09
Bug/bb 16040 update status when participants or limit changes

EnriqueSalazar:
When a task is full and the activity manager increases the number of people needed, the task detail page still shows that the task is full

https://onepercentclub.atlassian.net/browse/BB-16040

- BB-16068 Backend: status doesn't change when number of people needed increases
- BB-16069 Backend: Capacity not checked when participan is removed


eodolphi:
Nice!

PR #3926, opened 2019-12-10
Feature/bulk transitions

gannetson:


bluebottle/activities/admin.py
```diff
@@ -173,7 +173,7 @@ class ActivityAdmin(PolymorphicParentModelAdmin, FSMAdmin):
     list_filter = (PolymorphicChildModelFilter, ActivityStatusFilter, 'highlight')
     list_editable = ('highlight',)
 
-    list_display = ['__unicode__', 'created', 'type', 'combined_status',
```
eodolphi: Why do we stop doing here?

eodolphi: All activities need this right?

bluebottle/funding/transitions.py
```diff
@@ -25,12 +26,32 @@ def target_reached(self):
...
+            return errors
+
+    def is_valid(self):
+        errors = [
+            error.message[0] for error in self.instance.errors
+        ]
+
+        if errors:
+            return errors
+
```
eodolphi: Why do we not use the methods on the parent class?

gannetson: Because `self` will then expect a `ActitivityTransitions` instance, but a `FundingTransitions` will be provided. I guess there is a nicer way to do this, but I couldn't work it out.

PR #3939, opened 2019-12-13
Check stripe account status

gannetson:


eodolphi:
Do we want this to go to a release branch? Or will we test this on staging?

PR #3940, opened 2019-12-13
Connect account in funding serializer

gannetson:


eodolphi:
Why do we need this? In the frontend we get the list of account the user has and select the one for the funding on the fly.

gannetson:
The connect_account is in related for funding api call. Apparently @EnriqueSalazar had to do weird stuff to avoid race conditions when loading the connect_account through the bank_account from the funding activity. Adding connect_account also to the includes fixes this.

eodolphi:
I am closing this then. I am pretty sure he fixed it after he figured out how the route of the kyc-step works. This is not needed, and potentially slows down all requests with fundings.

PR #3941, opened 2019-12-13
Add links to event resource that allow users to add the event to their

eodolphi:
calendar.

- Add link to google calendar
- Add link to microsoft outlook
- Add link to ical download
- Add view that lets the user download an Ical file

BB-16087 #resolve
BB-16089 #resolve

### DIFF
--- a/bluebottle/events/models.py
+++ b/bluebottle/events/models.py
@@ -127,7 +127,8 @@ class Event(Activity):
     @property
     def google_calendar_link(self):
         def format_date(date):
-            return date.astimezone(utc).strftime('%Y%m%dT%H%M%SZ')
+            if date:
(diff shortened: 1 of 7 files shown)


bluebottle/events/models.py
```diff
@@ -118,6 +120,60 @@ def participants(self):
...
+            'enddt': format_date(self.end),
+            'body': '{}\n{}'.format(self.description, self.get_absolute_url()),
+        }
+
+        if self.location:
+            params['location'] = self.location.formatted_address
+
+        prepared_request.prepare_url(url, params)
+        return prepared_request.url
+
```
gannetson: Is there a nicer way then adding this to model? Probably fine for now though...

PR #3968, opened 2020-01-13
Hotfix/stripe account status

gannetson:


eodolphi:
Looks good. We do not need any frontend changes for this right?

gannetson:
Not entirely sure. I think it just worked like this, but we _are_ introducing a new status `incomplete` that the front-end doesn't handle yet. But since front-end will always ask Stripe for the details, I think this just might work (remember testing it).

Pieter-Rees:
Can we make a test release branch for this so that Charif can test this afternoon?

gannetson:
We can. It will overwrite the bulk transition release branch though

PR #3978, opened 2020-01-20
Do some optimizations:

eodolphi:
- Cache django money rates
- Cache resource permissions
- Make sure we do prefetch_related in filtered related field


eodolphi:
We need to think about this a bit more. In general we should not fetch all contributions. And we should fix how we do wallposts.

There should somehow be a paged interface that show all contributions and wallposts. And it should be possible to create wallposts for non-donation contributions.

eodolphi:
But this does more then half the time it takes to render a large funding resource.

PR #3995, opened 2020-01-29
Feature/more admin settings

gannetson:


eodolphi:
Are we skipping the migration for this, and do that by hand?

PR #4002, opened 2020-02-04
localize timezone

EnriqueSalazar:
[BB-16476 When campaign review state is approved, duration camping triggers a deadline +1 hour](https://onepercentclub.atlassian.net/browse/BB-16476)


eodolphi:
Looks good, but does need a test. Probably somewhere in `test_transitions.py`

eodolphi:
Could you look into creating a migration that corrects all current deadlines?

EnriqueSalazar:
> Could you look into creating a migration that corrects all current deadlines?

Done.

PR #4014, opened 2020-02-11
Hotfix/show refunded funding donations

eodolphi:


bluebottle/funding/transitions.py
```diff
@@ -222,7 +241,7 @@ class values(DjangoChoices):
         started = ChoiceItem('started', _('started'))
         succeeded = ChoiceItem('succeeded', _('succeeded'))
         failed = ChoiceItem('failed', _('failed'))
-        cancelled = ChoiceItem('cancelled', _('cancelled'))
+        cancelled = ChoiceItem('activity-refunded', _('cancelled'))
```
gannetson: For consistency use `activity_refunded` instead of `activity-refunded`?
Also refactor `cancelled` and `cancel()` into `activity_refunded`?

eodolphi: Nice catch. That should not have changed at all.

PR #4019, opened 2020-02-13
Override makemessages and don't do fuzzy translations

gannetson:


eodolphi:
Apparently Rolf never knew how fuzzy matching worked. It is really nice most of the time though. Since he now understands, and we fixed all the fuzzy translations, this can be closed.

PR #4023, opened 2020-02-18
Hotfix/event start time

eodolphi:


bluebottle/assignments/tests/test_models.py
```diff
@@ -55,6 +55,18 @@ def test_date_changed(self):
...
+            title='Test Title',
+            status='open',
+            end_date_type='on_date',
+            preparation=5,
+            date=now() + timedelta(days=4),
+        )
+
+        assignment.end_date_type = 'deadline'
+        assignment.save()
+
```
eodolphi: This should assert at least something. Without is it just tests that no exceptions are thrown.

EnriqueSalazar: assertions added

bluebottle/assignments/models.py
```diff
@@ -70,6 +88,19 @@ def required_fields(self):
...
+            tz_name = tf.timezone_at(
+                lng=self.location.position.x,
+                lat=self.location.position.y
+            )
+            tz = pytz.timezone(tz_name)
+
+            return self.date.astimezone(tz)
+        else:
+            return self.date
+
```
gannetson: So what will happen if you select 11:00 as time and then select Helsinki as location? Will then the `local_date` change to 12:00? 
What I would expect is that setting a location would change the timezone and thus changing the `date`, which still would be in UTC and then use this `local_date` to display the time in local format. 
In fact that _is_ how this method is working, but I didn't see the logic that changes the user input date to UTC based on timezone of the location.

bluebottle/assignments/serializers.py
```diff
@@ -27,13 +27,16 @@ class Meta(BaseActivityListSerializer.Meta):
...
             'end_date',
+            'date',
             'end_date_type',
             'registration_deadline',
             'capacity',
             'expertise',
             'duration',
             'location',
             'permissions',
+            'start_time',
```
gannetson: Through the admin you are only exposing `start` and here only `start_time`. What would happen if you change `start` in admin? You won't see those changes in the front-end, right?
Same for the other way around.

gannetson:
@EnriqueSalazar / @eodolphi In general there are too many changes in this PR which makes it hard to review. Maybe move the time zone changes to another PR, since that doesn't seem finished (unless we have that necessary changes in already another branch/master that I might have missed). It's not clear to me if we are now completely switched to using `Event.start` and `Event.date` here. Some parts still seem to use `start_time` or `end_date` or `end_time`. I guess there should be a reef PR to go with this too, but I couldn't find this. I'm just dicing into this so it might very well be that I'm just missing part of the changes, but please indulge ;-)

bluebottle/assignments/models.py
eodolphi: `self.date` is always in UTC. So this returns `self.date` in your local timezone. That is exactly what you want.

If a date is saved in the frontend, it will also be in the user's timezone (in the corresponding release branch), and will be saved to the db in UTC. So this will get you the time back as you inputted it (if the location is in your local timezone).

bluebottle/assignments/admin.py
```diff
@@ -120,8 +121,9 @@ class AssignmentAdmin(ActivityChildAdmin):
         ('initiative__title', 'Initiative'),
         ('expertise', 'Expertise'),
         ('end_date_type', 'End Time Type'),
-        ('end_date', 'End Date'),
+        ('end', 'End Date'),
```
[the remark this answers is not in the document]
eodolphi: Jep. I'll change that.

bluebottle/assignments/serializers.py
eodolphi: Start time could just go. I kept it in so that during the release, reads do not break. The admin should just expose date

bluebottle/assignments/serializers.py
```diff
@@ -72,6 +75,7 @@ class Meta(BaseActivitySerializer.Meta):
         fields = BaseActivitySerializer.Meta.fields + (
             'is_online',
             'end_date',
```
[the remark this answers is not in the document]
eodolphi: This was just so that during the release, reads won't break. I can also take it out.

PR #4025, opened 2020-02-18
Send a message to all followers of all approved activities when

eodolphi:
a wallpost is posted on an initiative

BB-16522 #resolve

### DIFF
--- a/bluebottle/initiatives/messages.py
+++ b/bluebottle/initiatives/messages.py
@@ -88,9 +88,14 @@ class InitiativeWallpostFollowerMessage(TransitionMessage):
 
     def get_recipients(self):
         initiative = self.obj.content_object
-        follows = initiative.follows.filter(
-            user__campaign_notifications=True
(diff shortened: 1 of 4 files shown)


bluebottle/initiatives/messages.py
```diff
@@ -88,9 +88,14 @@ class InitiativeWallpostFollowerMessage(TransitionMessage):
...
+        for activity in initiative.activities.filter(
+            review_status='approved'
+        ):
+            follows += activity.follows.filter(
+                user__campaign_notifications=True
+            ).exclude(
+                user__in=(self.obj.author, initiative.owner)
+            )
+
+        return set(follow.user for follow in follows)
```
gannetson: So I guess a user can't "follow" an Initiative, right? Can we then remove this from `initiatives/models.py`:
`follows = GenericRelation(Follow, object_id_field='instance_id')`

PR #4032, opened 2020-02-24
deletes inline deleted objects in admin

EnriqueSalazar:


gannetson:
Maybe a test to prove that this works might be nice. 
Maybe then we can check if changing `formset.save_m2m()` into `formset.save()` or `super(EventAdmin, self).save_formset(request, form, formset, change)` does the trick instead of adding that `for deletedObject in formset.deleted_objects:` bit.

gannetson:
Other whist change `deletedObject` to python style `deleted_object` ;-)

PR #4074, opened 2020-03-30
Use obscure filenames

gannetson:
BB-16732 #resolve


eodolphi:
This works, but does not fuly fix the issue. The files are still accesible. We have a subdirectory in 'private' or something  that is not served. Task documents should come from there.

PR #4078, opened 2020-03-31
response 401 when logout unauthenticated

EnriqueSalazar:
If applicable

### DIFF
--- a/bluebottle/bb_accounts/views.py
+++ b/bluebottle/bb_accounts/views.py
@@ -157,14 +157,14 @@ class Logout(generics.CreateAPIView):
     Log the user out
 
     """
+    permission_classes = (IsAuthenticated, )
+
     def create(self, request, *args, **kwargs):
         if self.request.user.is_authenticated:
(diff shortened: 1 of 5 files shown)


bluebottle/bb_accounts/views.py
```diff
@@ -162,7 +162,9 @@ def create(self, request, *args, **kwargs):
             self.request.user.last_logout = timezone.now()
             self.request.user.save()
 
-        return response.Response('', status=status.HTTP_204_NO_CONTENT)
+            return response.Response('', status=status.HTTP_204_NO_CONTENT)
+
+        return response.Response('', status=status.HTTP_401_UNAUTHORIZED)
```
eodolphi, line 167: This works, but it is probably a little clear to just add a correct permission class:

```
permission_classes = (IsAuthenticated, )
```

EnriqueSalazar: Ok done. looks much better.

PR #4115, opened 2020-05-14
Add assignment transitions

gannetson:
If applicable


bluebottle/assignments/models.py
```diff
@@ -100,12 +98,10 @@ def stats(self):
         contributions = self.contributions.instance_of(Applicant)
 
         stats = contributions.filter(
-            status=ApplicantTransitions.values.succeeded).\
+            status='succeeded').\
```
gannetson, line 101: Maybe do an import at the beginning of this method so we can do `ApplicantStateMachine.succeeded.value` here?
Or something like `Applicant.states.succeeded.value` or `self.applicants.__class__.states.succeeded.value`.
Would be nice to skip the literal strings here, right?

bluebottle/assignments/models.py
```diff
@@ -131,76 +127,61 @@ class JSONAPIMeta:
...
+    #     if self.status in [AssignmentTransitions.values.running,
+    #                        AssignmentTransitions.values.full,
+    #                        AssignmentTransitions.values.open]:
+    #         if len(self.accepted_applicants):
+    #             self.transitions.succeed()
+    #             self.save()
+    #         else:
+    #             self.transitions.expire()
+    #
+    #         self.save()
```
gannetson, line 173: If this is covered elsewhere, we should get rid of it.

PR #4131, opened 2020-05-29
Fix Payout states

gannetson:
If applicable

### DIFF
--- a/bluebottle/funding/views.py
+++ b/bluebottle/funding/views.py
@@ -184,9 +184,7 @@ class PayoutDetails(JsonApiViewMixin, AutoPrefetchMixin, RetrieveUpdateAPIView):
             serializer.instance.states.schedule()
         if status == 'started':
             serializer.instance.states.start()
-        if status == 'succeeded':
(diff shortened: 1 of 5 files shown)


bluebottle/funding/views.py
```diff
@@ -177,19 +177,23 @@ class PayoutDetails(JsonApiViewMixin, AutoPrefetchMixin, RetrieveUpdateAPIView):
...
+        if status == 'started':
+            serializer.instance.states.start()
         if status == 'succeeded':
-            payout.states.succeed(save=True)
-        if status == 'confirm':
-            payout.states.succeed(save=True)
+            serializer.instance.states.succeed()
+        if status == 'confirmed':
+            serializer.instance.states.succeed()
+        print status
```
eodolphi, line 195: This should not be there.

Maybe group the same call in the same if with `if status in ('new', 'scheduled', 're_scheduled', ):`

gannetson: It's informative!

PR #4201, opened 2020-08-17
Upgrade jquery

eodolphi:
If applicable


bluebottle/bluebottle_dashboard/static/jet/js/src/features/sidebar/main.js
```diff
@@ -3,7 +3,6 @@ var SideBarApplicationPinning = require('./application-pinning');
 var SideBarBookmarks = require('./bookmarks');
 var SideBarPopup = require('./popup');
 
-require('perfect-scrollbar/jquery')($);
```
gannetson, line 6: Is it right to remove this?

eodolphi: Fixed by installing the default package with the same version instead of the one from github.

PR #4205, opened 2020-08-17
Restrict mimetypes of files in admin

eodolphi:
If applicable

### DIFF
--- a/bluebottle/files/models.py
+++ b/bluebottle/files/models.py
@@ -1,6 +1,7 @@
 import os
 import uuid
 
+from django.conf import settings
 from django.contrib.contenttypes.fields import GenericForeignKey
 from django.contrib.contenttypes.models import ContentType
 from django.db import models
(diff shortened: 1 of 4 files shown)


bluebottle/files/models.py
```diff
@@ -9,12 +9,27 @@
...
 
 class File(AnonymizationMixin, models.Model):
     id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
     created = models.DateField(_('created'), default=timezone.now)
-    file = models.FileField(_('file'), upload_to='files')
+    file = models.FileField(
+        _('file'),
+        upload_to='files',
+        validators=[FileMimetypeValidator(
+            allowed_mimetypes=[
```
gannetson, line 22: Can we user `settings.PRIVATE_FILE_ALLOWED_MIME_TYPES` here?

eodolphi: Maybe. Is that list always the same? This includes zips, which might not be a good idea for private files.

eodolphi: Oh no, this is another thing. Yeah, that sound like a good idea.

bluebottle/utils/validators.py
```diff
@@ -75,3 +79,56 @@ def deconstruct(self):
...
+        if code is not None:
+            self.code = code
+
+    def __call__(self, value):
+        mimetype = mime.from_buffer(value.file.read(1000))
+        value.file.seek(0)
+
+        if self.allowed_mimetypes is not None and mimetype not in self.allowed_mimetypes:
+            import ipdb
+            ipdb.set_trace()
```
gannetson, line 106: Remove this

PR #4210, opened 2020-08-18
Hotfix/funding st fixes

gannetson:
If applicable

### DIFF
--- a/bluebottle/activities/states.py
+++ b/bluebottle/activities/states.py
@@ -166,7 +166,7 @@ class ActivityStateMachine(ModelStateMachine):
         ],
         needs_work,
         name=_('Restore'),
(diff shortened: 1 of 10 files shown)


bluebottle/activities/states.py
```diff
@@ -164,9 +164,9 @@ def is_owner(self, user):
             cancelled,
             deleted
         ],
-        draft,
+        needs_work,
         name=_('Restore'),
-        description=_('Restore the activity. The will mark the activity as draft again.'),
+        description=_('Restore the activity. The will mark the activity as needs_work again.'),
```
eodolphi, line 169: Replace `needs_work` in the description with `needs work`.

gannetson: Fair enough

PR #4215, opened 2020-08-19
Hotfix/transition admin fixes

eodolphi:


bluebottle/activities/models.py
```diff
@@ -170,9 +170,9 @@ def save(self, *args, **kwargs):
 
     def __unicode__(self):
         if self.user:
-            return unicode(_("Organizer: {name}".format(name=self.user.full_name)))
+            return self.user.full_name
         else:
-            return _('Organizer')
```
gannetson, line 175: This is here mainly for documentation purposes. If there is an organizer, it would always have a user. We might want to change how we make sure documentation prints nicely but now side effects might render like "Send message to None"

gannetson: Guess you fixed this with the templates...?

PR #4227, opened 2020-08-24
Hotfix/use state name instead of status

eodolphi:
If applicable


bluebottle/fsm/forms.py
```diff
@@ -22,6 +22,7 @@ def __new__(cls, name, bases, attrs):
                     required=False,
                     widget=TransitionSelectWidget()
                 )
+                print name, machine.field
```
gannetson, line 25: Remove

gannetson (approved):

PR #4242, opened 2020-08-28
Fix Mpesa push payments

gannetson:
If applicable


bluebottle/funding_lipisha/utils.py
```diff
@@ -151,56 +152,6 @@ def generate_error_response(reference):
     }
 
 
-def update_webhook_payment(data):
```
gannetson, line 154: We were not using this...

PR #4244, opened 2020-08-28
Never auto-submit funding activities. Also not when approving the

eodolphi:
initiative.

Fix effects when auto-submitting events.

If applicable

### DIFF
--- a/bluebottle/events/tests/test_api.py
+++ b/bluebottle/events/tests/test_api.py
@@ -581,8 +581,20 @@ class EventTransitionTestCase(BluebottleTestCase):
         data = json.loads(response.content)
         self.assertEqual(data['errors'][0], "Transition is not available")
 
-    def test_approve(self):
+    def test_reject(self):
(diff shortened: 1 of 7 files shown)


bluebottle/events/tests/test_api.py
```diff
@@ -582,7 +582,7 @@ def test_delete_by_other_user(self):
         self.assertEqual(data['errors'][0], "Transition is not available")
 
     def test_approve(self):
-        self.review_data['data']['attributes']['transition'] = 'approve'
+        self.review_data['data']['attributes']['transition'] = 'reject'
```
gannetson, line 585: Hmmm not sure about this. Shouldn't we test that you can or cannot use the approve transition?
Now at least the method name should change....

eodolphi: I kind of see what you mean. Also test that a transition that does not exist correctly works.
I'll have a quick look.

PR #4253, opened 2020-08-31
Feature/add impact to export

gannetson:
If applicable


eodolphi:
Doesn't this need a settings change to actually export things?

eodolphi (approved):

PR #4260, opened 2020-09-02
Feature/pass request user

gannetson:
If applicable


bluebottle/fsm/triggers.py
```diff
@@ -58,6 +58,7 @@ class TriggerMixin(object):
     periodic_tasks = []
 
     def __init__(self, *args, **kwargs):
+        self._request_user = kwargs.pop('request_user', None)
```
gannetson, line 61: This is not OK.

PR #4300, opened 2020-09-28
Add theme detail api

gannetson:

### DIFF
--- a/bluebottle/initiatives/tests/test_api.py
+++ b/bluebottle/initiatives/tests/test_api.py
@@ -1116,8 +1116,7 @@ class ThemeAPITestCase(BluebottleTestCase):
         super(ThemeAPITestCase, self).setUp()
 
         self.client = JSONAPITestClient()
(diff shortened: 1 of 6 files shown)


eodolphi:
We should also do this for skill right?

bluebottle/initiatives/tests/test_api.py
```diff
@@ -1138,6 +1138,14 @@ def test_list(self):
...
         self.assertEqual(theme.name, result['attributes']['name'])
 
+    def test_detail(self):
+        theme = self.themes[0]
+        url = reverse('initiative-theme', args=(theme.id,))
+        response = self.client.get(url, user=self.user)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        result = response.json()['data']
+        self.assertEqual(result['attributes']['name'], theme.name)
+
```
eodolphi, line 1148: Can we add tests that also explicitly test the permissions?

PR #4330, opened 2020-10-12
Add a signal that updates the segments of all open activities if a

eodolphi:
segment is added or removed from a user

BB-17549 #resolve

If applicable

### DIFF
--- a/bluebottle/activities/tests/test_models.py
+++ b/bluebottle/activities/tests/test_models.py
@@ -28,7 +28,7 @@ class ActivitySegmentsTestCase(TestCase):
         self.assertTrue(self.team in activity.segments.all())
 
     def test_segments_already_set(self):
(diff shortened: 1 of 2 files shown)


bluebottle/activities/tests/test_models.py
```diff
@@ -37,3 +37,33 @@ def test_segments_already_set(self):
         self.assertTrue(self.unit in activity.segments.all())
         self.assertTrue(self.team in activity.segments.all())
         self.assertFalse(self.other_team in activity.segments.all())
+
+    def test_segments_already_set_open(self):
+        activity = EventFactory.create(owner=self.user, status='open')
+        self.user.segments.remove(self.team)
+        self.user.segments.add(self.other_team)
+
+        activity.save()
```
gannetson, line 46: This should not be needed, right?

gannetson (approved):

PR #4333, opened 2020-10-13
Feature/online meeting url

eodolphi:
If applicable


bluebottle/events/models.py
```diff
@@ -135,6 +136,19 @@ def participants(self):
...
+    @property
+    def details(self):
+        details = HTMLParser().unescape(
+            u'{}\n{}'.format(
+                strip_tags(self.description), self.get_absolute_url()
+            )
+        )
+
+        if self.is_online and self.online_meeting_url:
+            details += _('\nJoin: {}').format(self.online_meeting_url)
```
gannetson, line 148: I think it's better to have named placeholders in translatable strings:
`_('\nJoin: {online_meeting_url}').format(online_meeting_url=self.online_meeting_url)`

eodolphi: Having named placeholders if there is only one, only makes it easier to mess up the translation.

gannetson (approved):

PR #4358, opened 2020-11-02
Add test but fail to fix

gannetson:
If applicable


bluebottle/bluebottle_dashboard/urls.py
```diff
@@ -0,0 +1,66 @@
+import django
```
eodolphi, line 1: Why is this here? The monkey patching should be enough right? And why do we mark these urls as `csrf_exempt`?

gannetson: Remove this file. We are not using it

gannetson: We are not using this file. Please remove it.

PR #4409, opened 2020-12-03
S&T descriptions

gannetson:
If applicable

### DIFF
--- a/bluebottle/time_based/messages.py
+++ b/bluebottle/time_based/messages.py
@@ -101,11 +101,8 @@ class ParticipantAddedNotification(TransitionMessage):
     }
 
     def get_recipients(self):
-        """activity owner"""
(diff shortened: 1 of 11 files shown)


bluebottle/time_based/messages.py
```diff
@@ -101,11 +101,8 @@ class ParticipantAddedNotification(TransitionMessage):
     }
 
     def get_recipients(self):
-        """activity owner"""
-        if self.options.get('user') != self.obj.user:
-            return [self.obj.user]
-        else:
-            return []
+        """participant"""
```
eodolphi, line 104: I am pretty sure we want this here. We should only send this message if someone else adds you to an activity.

eodolphi (changes requested):

eodolphi (approved):

PR #4421, opened 2020-12-09
Always include initiative location

gannetson:
If applicable


eodolphi (changes requested):
Looks like you need to rebase. This includes all the time based activity changes.

eodolphi:
This could do with a test, but lets fix that when we remove the old view.

PR #4432, opened 2020-12-16
Fix rendering of timecontribution transitions.

eodolphi:
Fix succeeding / failing and resetting of time contributions on various
transtions.
Fix creating and succeeding contributions on period participants.

If applicable

### DIFF
--- a/bluebottle/time_based/tasks.py
+++ b/bluebottle/time_based/tasks.py
@@ -14,7 +14,7 @@ logger = logging.getLogger('bluebottle')
 
 
 @periodic_task(
-    run_every=(crontab(minute='*')),
+    run_every=(crontab(minute='*/15')),
     name="on_a_date_tasks",
     ignore_result=True
 )
@@ -26,7 +26,7 @@ def on_a_date_tasks():
 
 
 @periodic_task(
-    run_every=(crontab(minute='*')),
+    run_every=(crontab(minute='*/15')),
     name="with_a_deadline_tasks",
     ignore_result=True
 )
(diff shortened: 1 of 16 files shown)


bluebottle/time_based/tasks.py
```diff
@@ -14,7 +14,7 @@
 
 
 @periodic_task(
-    run_every=(crontab(minute='*/15')),
+    run_every=(crontab(minute='*')),
```
gannetson, line 17: Is this for testing, or do we think it's not heavy to run it this often?

bluebottle/time_based/tests/test_triggers.py
```diff
@@ -746,58 +779,14 @@ def test_no_review_succeed_after_cancel(self):
...
-            self.participants[0].contributions.get().status,
-            'failed'
-        )
-
-    def test_reaccept(self):
-        super().test_reaccept()
-
-        self.assertEqual(
-            self.participants[0].contributions.get().status,
-            'new'
```
gannetson, line 800: Don't we want these?

gannetson (changes requested):

bluebottle/time_based/tests/test_triggers.py
eodolphi: I moved those up to the parent tests.

bluebottle/time_based/tasks.py
eodolphi: That was for testing. I'll revert that.

PR #4440, opened 2020-12-21
Add promoter to activity api

gannetson:
We can add this after time-based release.


eodolphi:
Does this only add the promotor to the details view? If not, also add a preselect related.

gannetson:
> Does this only add the promotor to the details view? If not, also add a preselect related.

Yes I think it's only the detail serializers.

Issue #4313, opened 2020-10-05
Django 3


jayvdb:
I see there is lots of work going on for Python 3 support at https://github.com/onepercentclub/bluebottle/pull/4292

I tried my hand at leveraging off that to see if I can get Django 3 working, and am mostly successful in getting it operational enough to explore the app locally.

Latest CI result at https://travis-ci.org/github/jayvdb/bluebottle/jobs/732862786
Summary: of 1228 tests, 225 failures & 184 errors.

A lot of the errors are repeatitive, and most are due to the dependencies being pinned to Django 1.x compatible versions.  I'm trying to unpin only dependencies that really need unpinning, so-as to reduce the variability between Django 1.x and Django 3.x versions of bluebottle.

## Other dependent PRs / branches
- https://github.com/onepercentclub/bluebottle/pull/3387 Remove SaaS service salesforce (this isnt on any branch; it disappeared?)
- https://github.com/onepercentclub/django-tenant-extras/pull/60 - Django 2 & 3 support with tests
- https://github.com/django-fluent/django-fluent-contents/pull/107 - sort-thumbnail induced bug
- https://github.com/philippbosch/django-geoposition/pull/110 - Django 2 & 3 support
- https://github.com/jayvdb/django-tenant-schemas/commits/migrate-fix (hack around conflict in `--skip-checks`)
- https://github.com/jayvdb/bluebottle/commits/dj3-v2 - 60+ commits and counting, of varying messiness - most notably removing django-jet, which turned into a SaaS service for Django 2+.  IMO their licensing fees are not money well spent, as similar exists in the OSS world.

##  Why Django 3?
1. Because most dependency projects are already dropping Django 2, so if there are problems with porting to Django 2 it would be harder to get help,
2. All top-tier distros have dropped Django 2 from their current or pending release, and
3. I expect the Django 3 effort to take a while to stablise and before existing bluebottle production systems would be ready to switch to Django 3, and by then Django 2 will be even more dead.
4. Creating migration across major Django versions is quite a bit of effort.  May as well skip one.

Obviously most of this needs to wait until the Python 3 work lands, but I'd like to know the best way to structure this in order to get favourable reviews to get things merged early.

[...]

PR #4454, opened 2021-01-01
Django 2+

jayvdb:
The simpler part of the commit series described at https://github.com/onepercentclub/bluebottle/issues/4313


jayvdb:
Wonderful.  I'll work on the next batch of commits for #4313

PR #4473, opened 2021-01-13
Hotfix/slot participant api

eodolphi:
If applicable


bluebottle/time_based/effects.py
```diff
@@ -197,3 +197,19 @@ def post_save(self, **kwargs):
...
+
+    @property
+    def display(self):
+        return self.instance.activity.slot_selection == 'all'
+
+    def post_save(self, **kwargs):
+        slot = self.instance
+        activity = self.instance.activity
+        if activity.slot_selection == 'all':
+            for participant in activity.accepted_participants:
```
gannetson, line 214: I think we now create slot_participants on participant.initiate, so we should either move that to participant.approve or add slot_participants for all acitivity.participants.
In the first case we should also have some logic to the participant.approve/restore that loops over the slots and do a get_or_create or something similar.

eodolphi: Yeah, you are right. I wanted to fix that but forgot about that. There is a failing test for this though.

PR #4493, opened 2021-01-25
Feature/participant admin

gannetson:
If applicable


bluebottle/time_based/admin.py
```diff
@@ -404,10 +415,122 @@ class TimeContributionAdmin(ContributionChildAdmin):
...
+class SlotWidget(TextInput):
+
+    template_name = 'admin/widgets/slot_widget.html'
+
+
+class ParticipantSlotForm(ModelForm):
+    checked = BooleanField(label=_('Participating'), initial=True, required=False)
+
+    def __init__(self, *args, **kwargs):
+        super(ParticipantSlotForm, self).__init__(*args, **kwargs)
```
eodolphi, line 427: Why do you use strings as default here. Isn'nt None much more clear?

bluebottle/time_based/admin.py
```diff
@@ -404,10 +415,122 @@ class TimeContributionAdmin(ContributionChildAdmin):
...
+            for slot in instance.activity.slots.exclude(slot_participants__participant=instance).all():
+                new.append({
+                    'slot': slot,
+                    'checked': False
+                })
+
+            kwargs.update({'initial': new})
+        super(ParticipantSlotFormSet, self).__init__(*args, **kwargs)
+
+    def save_new_objects(self, commit=True):
```
eodolphi, line 465: Please specifiy what is overridden. 

Wasn't it the plan to transition unchecked objects instead of deleting them?

bluebottle/fsm/admin.py
```diff
@@ -65,8 +65,9 @@ def changeform_view(self, request, object_id=None, form_url='', extra_context=No
             for formset in formsets:
                 for form in formset:
                     if isinstance(form.instance, TriggerMixin):
-                        self.save_form(request, form, change=True)
-                        effects += form.instance.execute_triggers(user=request.user, send_messages=send_messages)
+                        instance = self.save_form(request, form, change=True)
+                        if instance:
+                            effects += form.instance.execute_triggers(user=request.user, send_messages=send_messages)
```
eodolphi, line 70: Either use the returned instance or the instance on the form, but not both.

bluebottle/time_based/admin.py
```diff
@@ -404,10 +415,122 @@ class TimeContributionAdmin(ContributionChildAdmin):
...
+
+    def __init__(self, *args, **kwargs):
+        super(ParticipantSlotForm, self).__init__(*args, **kwargs)
+        initial = kwargs.get('initial', '')
+        slot = ''
+        if initial:
+            slot = initial['slot']
+        instance = kwargs.get('instance', '')
+        if instance:
+            slot = instance.slot
```
eodolphi, line 434: This might set the slot to `''`, which is wrong.

PR #4499, opened 2021-01-28
Feature/more tests

gannetson:
If applicable


bluebottle/time_based/models.py
```diff
@@ -375,7 +375,6 @@ class DateActivitySlot(ActivitySlot):
     @property
     def required_fields(self):
         fields = [
-            'is_online',
             'start',
```
eodolphi, line 378: This field should be required on the model. You need to make a choice here before the activity can go live.

eodolphi (changes requested):

bluebottle/time_based/models.py
gannetson: Nope. This is only needed on slot level. We should remove this all the way later on.

PR #4501, opened 2021-01-29
Hotfix/duplicating location

gannetson:
If applicable


bluebottle/activities/admin.py
```diff
@@ -508,7 +508,7 @@ def location_link(self, obj):
     def get_list_display(self, request):
         fields = self.list_display
         from bluebottle.geo.models import Location
-        if Location.objects.count() and 'location' not in fields:
+        if Location.objects.count() and 'location_link' not in fields:
             fields += ['location_link']
```
eodolphi: To make sure this really does not happen, do not modify the list, but create a new list:
`fields = fields + ['location_link']`

eodolphi: Then you can also get rid of the `in` check.

eodolphi (approved):

eodolphi (approved):

PR #4506, opened 2021-02-02
Feature/preparation time

gannetson:
If applicable


bluebottle/time_based/admin.py
```diff
@@ -23,10 +23,34 @@
...
+            return super().get_fields(request, obj)
+        else:
+            return ['disabled']
+
+    def get_template(self):
+        pass
+
+    def disabled(self, obj):
+        return format_html('<i>{}</i>', obj)
+    disabled.short_description = _('First complete and submit the activity before managing participants.')
```
gannetson, line 42: Granted, this is probably not the cleanest way to show a message like this, but it does the trick. Otherwise we have to override formset or somehow have conditional template.

eodolphi (approved):

PR #4509, opened 2021-02-03
Fix calender links. Only add slots that user is participant of to ica…

eodolphi:
…l. Fixes for includes

If applicable


bluebottle/time_based/serializers.py
```diff
@@ -175,7 +169,9 @@ def get_my_contributor(self, instance):
     class Meta(TimeBasedBaseSerializer.Meta):
         model = DateActivity
         fields = TimeBasedBaseSerializer.Meta.fields + (
-            'utc_offset',
+            'start',
+            'duration',
+            'online_meeting_url',
```
gannetson, line 174: These just got removed from DateActivity

gannetson (approved):

PR #4529, opened 2021-02-16
Add participant export url

gannetson:
BB-18091 #resolve

If applicable


eodolphi (changes requested):
I guess we should also do this for period activities right?

gannetson:
> I guess we should also do this for period activities right?

You guessed right :prize:

eodolphi (approved):

PR #4530, opened 2021-02-16
Add title to slots

gannetson:
If applicable


bluebottle/time_based/admin.py
```diff
@@ -161,15 +161,14 @@ class DateActivityASlotInline(admin.TabularInline):
         },
     }
 
-    raw_id_fields = ['location']
```
eodolphi, line 164: This should be in the 1.2 release right?

eodolphi (changes requested):

PR #4587, opened 2021-03-17
Use random images for activities/intiatives if they aren't found (te…

gannetson:
…sting only)

If applicable


gannetson:
image

eodolphi (approved):

PR #4630, opened 2021-04-20
Hotfix/matching fixes

eodolphi:
If applicable


bluebottle/activities/messages.py
```diff
@@ -220,6 +220,11 @@ def get_activity_context(self, activity):
 
     def get_context(self, recipient, activities=None):
         context = super().get_context(recipient)
+        context['profile_incomplete'] = (
+            not (len(recipient.favourite_themes.all())) or
```
gannetson, line 224: `len(recipient.favourite_themes.all())` is probably almost the same as `recipient.favourite_themes.count()`  or `recipients.favourite_themes.exists()` right?

gannetson (approved):

bluebottle/activities/messages.py
eodolphi: Yes. `.count` is the same as `len()`. The last one is a bit more flexible since it also works on non-quersets .

PR #4688, opened 2021-05-31
Make user creation and login case insensitive

eodolphi:
If applicable


bluebottle/funding/effects.py
```diff
@@ -22,7 +22,10 @@ class GeneratePayoutsEffect(Effect):
 
     def post_save(self, **kwargs):
         from bluebottle.funding.models import Payout
-        Payout.generate(self.instance)
+        try:
+            Payout.generate(self.instance)
+        except AssertionError:
+            pass
```
gannetson, line 28: Why this?

bluebottle/members/tests/test_api.py
```diff
@@ -291,6 +314,42 @@ def test_create_twice(self):
...
+        self.assertEqual(
+            response.json()['non_field_errors'][0]['type'],
+            'email'
+        )
+
+        self.assertEqual(
+            response.json()['non_field_errors'][0]['id'],
+            user_id
+        )
+
```
gannetson, line 352: Maybe another test to specifically test the flow where user uses email sign-up first and then Facebook or v.v.? I think that is most common scenario. Or will that be using the same UserCreateSerializer?

scripts/merge_duplicate_users.py
```diff
@@ -0,0 +1,38 @@
...
+
+                for contributor in Contributor.objects.filter(user=duplicate):
+                    contributor.user = first
+                    contributor.execute_triggers(send_messages=False)
+                    contributor.save()
+
+                for initiative in Initiative.objects.filter(owner=duplicate):
+                    initiative.owner = first
+                    initiative.execute_triggers(send_messages=False)
+                    initiative.save()
```
gannetson, line 36: Are we skipping all other user related objects? Wall posts, user activity, permissions, themes, skills, avatar etc? Not sure how big this issue is though.

gannetson: Is it a wild guess that the first sign-up is the account that is used the most or did you investigate this? Which one has the most activity? Again, not sure how much effort we should put into this. Maybe keep some reference to the one user account we picked in the anonymised one? Abuse the about_me field to print 'duplicate of #id'? Then we can always repair it afterwards..
.

gannetson: or `duplicate.email = '{}-duplo-{}'.format(first.email, duplicate.id)`
Needs another save though...

bluebottle/members/tests/test_api.py
eodolphi: Facebook login is case-insensitive already. But is really hard to test from our code. There are tests in social-django though that verify this.

bluebottle/funding/effects.py
eodolphi: There was at least one old activity that changed status because of the save. This triggered this effect, however there already was a payout generated for this activity. This makes sures that the effect still works even if the check fails.

scripts/merge_duplicate_users.py
eodolphi: > Are we skipping all other user related objects? Wall posts, user activity, permissions, themes, skills, avatar etc? Not sure how big this issue is though.

I was not exactly sure what other things to migrate. Wallpost seems logical. Permissions not so much since admin accounts should not be part of this, and neither should partner accounts. 

Themes, skills and avatars I do not really know. I do not think there will be many of those.

eodolphi: > or `duplicate.email = '{}-duplo-{}'.format(first.email, duplicate.id)`
> Needs another save though...

Jep. I added a way to identify merged users now.

eodolphi: > Is it a wild guess that the first sign-up is the account that is used the most or did you investigate this? Which one has the most activity? Again, not sure how much effort we should put into this. Maybe keep some reference to the one user account we picked in the anonymised one? Abuse the about_me field to print 'duplicate of #id'? Then we can always repair it afterwards..
> .

Yes. I did not find a single account that had more activity on the newer account. Most accounts had no activity on the second. In the end it does not really matter though. As long as the older stuff is merged.

PR #4711, opened 2021-06-25
Split contribution create effect. Make sure we do not create period

eodolphi:
contributions for non-periodic activities.

BB-18780 #resolve

If applicable

### DIFF
--- a/bluebottle/time_based/effects.py
+++ b/bluebottle/time_based/effects.py
@@ -46,6 +46,34 @@ class CreatePreparationTimeContributionEffect(Effect):
             contribution.save()
 
 
+class CreateOverallTimeContributionEffect(Effect):
+    title = _('Create contribution')
+    template = 'admin/create_period_time_contribution.html'
+
+    def post_save(self, **kwargs):
+        tz = get_current_timezone()
+        activity = self.instance.activity
+
+        # Just create a contribution for the full period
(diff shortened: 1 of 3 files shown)


bluebottle/time_based/effects.py
```diff
@@ -46,6 +46,34 @@ def post_save(self, **kwargs):
...
+        start = activity.start or date.today()
+        end = activity.deadline if hasattr(activity, 'deadline') else None
+
+        TimeContribution.objects.create(
+            contributor=self.instance,
+            contribution_type=ContributionTypeChoices.period,
+            value=activity.duration,
+            start=tz.localize(datetime.combine(start, datetime.min.time())),
+            end=tz.localize(datetime.combine(end, datetime.min.time())) if end else None,
+        )
```
gannetson, line 67: It looks like this bit of code was the same as before, so it doesn't fix the actual problem. Or is there something I missed?  Or is it enough that this bit was removed from `CreatePeriodTimeContributionEffect`, so it won't get called by the celery task?

eodolphi: Exactly. This will not be called in  a periodictask anymore, only in initate.

eodolphi: There is a test for this.

gannetson (approved):

PR #4757, opened 2021-09-01
Validate inline form before saving

gannetson:
If applicable


eodolphi (approved):
Nice!

PR #4759, opened 2021-09-01
Add nicer arrayfield widget and test segments

gannetson:
If applicable


bluebottle/token_auth/tests/test_saml.py
```diff
@@ -1,4 +1,7 @@
 from future import standard_library
+
+from ...segments.tests.factories import SegmentTypeFactory, SegmentFactory
```
eodolphi, line 3: These always confuse me. Can we do full imports from bluebottle?

eodolphi (approved):

PR #4794, opened 2021-10-13
Change user address (place) + tests

gannetson:


bluebottle/members/migrations/0048_auto_20211012_1256.py
```diff
@@ -0,0 +1,33 @@
...
+
+# Generated by Django 2.2.24 on 2021-10-12 10:43
+
+from django.db import migrations, connection
+
+from bluebottle.clients.models import Client
+from bluebottle.clients.utils import LocalTenant
+
+
+def migrate_addresses(apps, schema_editor):
```
eodolphi, line 11: I think the tenant is already set.

eodolphi (approved):

PR #4820, opened 2021-11-02
Feature/overall collect contribution

eodolphi:
If applicable


bluebottle/collect/serializers.py
```diff
@@ -47,7 +47,7 @@ class CollectActivitySerializer(BaseActivitySerializer):
     def get_contributors(self, instance):
         user = self.context['request'].user
         return [
-            contributor for contributor in instance.contributors.all() if (
+            contributor for contributor in instance.contributors.exclude(user=None) if (
                 isinstance(contributor, CollectContributor) and (
                     contributor.status in [
                         CollectContributorStateMachine.new.value,
```
gannetson, line 53: We can do this as a filter on queryset rather than getting all items and then filter on iter, right?

gannetson (approved):

PR #4898, opened 2022-01-19
Migrate custom fields to segments

gannetson:
This will also change how you set segments in back office

Screenshot 2022-01-19 at 13 01 37


bluebottle/members/admin.py
```diff
@@ -346,20 +340,20 @@ def get_fieldsets(self, request, obj=None):
             if not PaymentProvider.objects.filter(Q(instance_of=PledgePaymentProvider)).count():
                 fieldsets[2][1]['fields'].remove('can_pledge')
 
-            if CustomMemberFieldSettings.objects.count():
+            if obj and (obj.is_staff or obj.is_superuser):
```
eodolphi, line 343: The `is_staff` should not matter here right, since if you can access this then you are always a staff user.

eodolphi (approved):

PR #4942, opened 2022-02-15
Demo/paginate contributors

gannetson:


bluebottle/utils/views.py
```diff
@@ -1,6 +1,6 @@
-from builtins import object
 import mimetypes
 import os
+from builtins import object
```
eodolphi, line 3: Instead of importing this class, just remove the `(object)` from the class definition.

eodolphi (approved):

PR #4959, opened 2022-02-23
added test to prove that a user does not have a location when a locat…

MomchilovP:


bluebottle/members/tests/test_admin.py
```diff
@@ -301,13 +301,15 @@ class MemberPlatformSettingsAdminTestCase(BluebottleAdminTestCase):
...
         self.app.set_user(self.superuser)
+        user = BlueBottleUserFactory.create()
         page = self.app.get(reverse('admin:members_memberplatformsettings_change'))
         form = page.forms[0]
         form['require_office'].checked = True
 
         form.submit()
         settings_platform = MemberPlatformSettings.load()
-        self.assertTrue(settings_platform.require_office)
+        if user.place is None:
```
eodolphi, line 311: Putting if statements in tests like this does not make any sense. Either `user.place` is always `None` in this test, and the line after this is never hit, or it is never `None` and this line is always hit. Test will, and should, always behave the same. If `if` statement would do something, sometimes the tests might fail, and sometimes they will not. That would be really annoying.

bluebottle/members/tests/test_admin.py
```diff
@@ -301,13 +301,15 @@ class MemberPlatformSettingsAdminTestCase(BluebottleAdminTestCase):
...
+            self.assertTrue(settings_platform.require_office)
```
eodolphi, line 313: Why would you expect that `settings_platform.require_office` would be something different if a certain user would exist? That is not at all what this test was supposed to verify. If this would have been the case, something would be really wrong.

bluebottle/members/tests/test_admin.py
eodolphi, line 311: How did you verify that this test actually tests anything? Always make sure that you see a newly written test fail. Either by writing the test, letting it fail and writing the code that fixes the test. Or, if you write a test for something that is already working, by termporarily breaking the code. Then run the test, see it fail, unbreak the code, run the test and see it pass.

There is no way that this test would ever break on the assertion you added.

eodolphi (changes requested):

PR #4964, opened 2022-02-24
Ticket/bb 19643 test locations

MomchilovP:
Tests for working cache are not done yet, tests for long/lat omitted need to be refined

### DIFF
--- a/bluebottle/initiatives/tests/test_api.py
+++ b/bluebottle/initiatives/tests/test_api.py
@@ -1525,12 +1525,13 @@ class InitiativeMapListTestCase(BluebottleTestCase):
         super(InitiativeMapListTestCase, self).setUp()
 
         self.url = reverse('initiative-map-list')
-        self.client = JSONAPITestClient()
 
     def test_list_initiatives(self):
         InitiativeFactory.create_batch(5, status='approved')
         response = self.client.get(
             self.url
         )
+        data = response.json()
 
         self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertTrue(len(data), 5)
--- a/bluebottle/initiatives/tests/test_api.py
+++ b/bluebottle/initiatives/tests/test_api.py
@@ -120,7 +120,7 @@ class InitiativeListAPITestCase(InitiativeAPITestCase):
         self.assertEqual(response_data['data']['attributes']['title'], ':)')
         self.assertNotEqual(response_data['data']['attributes']['slug'], '')
 
-    def test_create_missing_iamge(self):
+    def test_create_missing_image(self):
         data = {
             'data': {
                 'type': 'initiatives',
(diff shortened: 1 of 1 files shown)


bluebottle/initiatives/tests/test_api.py
```diff
@@ -1552,3 +1553,66 @@ def test_submitted_initiatives_dont_show_on_map(self):
...
+
+        lat = data[0]['position'][0]
+        long = data[0]['position'][1]
+        lat_2 = data[1]['position'][0]
+        long_2 = data[1]['position'][1]
+        self.assertTrue(lat in data[0]['position'])
+        self.assertTrue(long in data[0]['position'])
+        self.assertTrue(lat_2 in data[1]['position'])
+        self.assertTrue(long_2 in data[1]['position'])
+
```
gannetson, line 1580: You're doing this:
```
fruit = 'apple'
self.assertTrue(fruit in  ['apple'])
```
You probably want to check that `lat` or `data[0]['position'][0]` has a specific value., like
```
self.assertEqual(lat, '43.0579025')
```

bluebottle/initiatives/tests/test_api.py
```diff
@@ -1552,3 +1553,66 @@ def test_submitted_initiatives_dont_show_on_map(self):
...
+            place=GeolocationFactory(position=Point(23.6851594, 0))
+        )
+        InitiativeFactory(
+            status='approved',
+            place=GeolocationFactory(position=Point(0, 34.0579025))
+        )
+        InitiativeFactory(
+            status='approved',
+            place=GeolocationFactory(position=Point(23.6851594, 34.0579025))
+        )
```
gannetson, line 1593: You should create a position with `Point(0, 0)`, because that's the one that will be filtered out.

bluebottle/initiatives/tests/test_api.py
```diff
@@ -1552,3 +1553,66 @@ def test_submitted_initiatives_dont_show_on_map(self):
...
+
+        response = self.client.get(
+            self.url
+        )
+        data = response.json()
+
+        for i in range(len(data)):
+            if data[i]['position'][0] == 0.0:
+                del data[i]
+                break
```
gannetson, line 1603: No idea what you are doing here...

bluebottle/initiatives/tests/test_api.py
```diff
@@ -1552,3 +1553,66 @@ def test_submitted_initiatives_dont_show_on_map(self):
...
+        InitiativeFactory.create_batch(5, status='approved')
+        response = self.client.get(
+            self.url
+        )
+
+        data = response.json()
+        self.assertTrue(len(data), 5)
+
+        InitiativeFactory.create(status='approved')
+        self.assertTrue(len(data), 7)
```
gannetson, line 1615: 1. Make sure you cache is working for tests 
2. After creating an extra initiative you should `self.client.get` again, otherwise you're just looking at old data
3. Then assert you're getting the cached data so `self.assertTrue(len(data), 7)`
Probably step 1 is the hardest and you'll need further assistance. I would have to look through code or google this too.

gannetson (changes requested):

bluebottle/initiatives/tests/test_api.py
MomchilovP: No need to google it, I know how to make the cache work.

bluebottle/initiatives/tests/test_api.py
MomchilovP: Removing the initiatives that have a lat that is 0

gannetson: But why? You want to test the result from the response, not changing the data here in a test.

gannetson (approved):

PR #4970, opened 2022-02-28
Test date activity participant count

MomchilovP:


bluebottle/time_based/tests/test_admin.py
```diff
@@ -73,9 +73,24 @@ def test_list_activities_office(self):
...
+        DateParticipantFactory.create_batch(5, activity=activity)
+        url = reverse('admin:time_based_dateactivity_changelist')
+        response = self.app.get(url)
+        self.assertEqual(int(response.html.find_all('td')[5].get_text()), 5)
+
+    def test_start_date(self):
+        activity = DateActivityFactory.create()
+        DateActivitySlotFactory.create(activity=activity)
+        date = defaultfilters.date(activity.slots.all()[0].start) + ', '\
+               + defaultfilters.time(activity.slots.all()[0].start + timedelta(hours=2)) # noqa
```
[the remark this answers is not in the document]
eodolphi: In this test you want to verify that the earliest slot is shown. But since you created an activity with a single slot. This test does not verify that at all. You will need to create at least 2 slots, and then make sure that the earliest is displayed. Either by creating a slot that you now is earlier then other slots. Or by sorting the slots and taking the first.

eodolphi (changes requested):

bluebottle/time_based/tests/test_admin.py
```diff
@@ -73,9 +73,26 @@ def test_list_activities_office(self):
...
+        DateParticipantFactory.create_batch(5, activity=activity)
+        url = reverse('admin:time_based_dateactivity_changelist')
+        response = self.app.get(url)
+        self.assertEqual(int(response.html.find_all('td')[5].get_text()), 5)
 
-class DateActivityAdminScenarioTestCase(BluebottleAdminTestCase):
+    def test_start_date(self):
+        activity = DateActivityFactory.create()
+        DateActivitySlotFactory.create_batch(2, activity=activity)
+        start_time = activity.slots.all()[0].start.astimezone(get_current_timezone())
```
eodolphi, line 86: This looks a little fragile to me. The DateActivitySlot model has an ordering of `['start', 'id']`, so by default the slots are ordered by start date. But this is kind of an implementation detail, which is why we sort the slots explicitly in the admin property. Either add a comment that this test works because of this ordering, or explicitly find the earliest slot.

eodolphi (changes requested):

bluebottle/time_based/tests/test_admin.py
```diff
@@ -73,9 +73,42 @@ def test_list_activities_office(self):
...
+            activity=activity,
+            start=time_2
+        )
+        DateActivitySlotFactory.create(
+            activity=activity,
+            start=time_3
+        )
+
+        start_time = activity.slots.all()[0].start.astimezone(get_current_timezone())
+        self.assertTrue(start_time == time_2)
```
gannetson, line 103: This would be nicer
`self.assertEqual(start_time, time_2)`

bluebottle/time_based/tests/test_admin.py
```diff
@@ -73,9 +73,42 @@ def test_list_activities_office(self):
...
+
+        url = reverse('admin:time_based_dateactivity_changelist')
+        response = self.app.get(url)
+        slot_time = f'{defaultfilters.date(start_time)}, {defaultfilters.time(start_time)}'
```
gannetson, line 107: Maybe use `time_2` here instead. It's a bit more explicit.

bluebottle/time_based/tests/test_admin.py
MomchilovP: True, changed it 👌🏽

PR #4983, opened 2022-03-09
tested the donation date

MomchilovP:


bluebottle/funding/tests/test_admin.py
```diff
@@ -153,6 +154,18 @@ def test_donation_reward(self):
...
 
+    def test_contribution_date(self):
+        donation = DonorFactory.create(
+            activity=self.funding,
+            amount=Money(100, 'EUR')
+        )
+        donation.states.succeed(save=True)
+        self.client.force_login(self.superuser)
+        response = self.client.get(self.admin_url)
+        curr_time = now()
```
gannetson, line 165: You probably want to use this as `donation.created` too.

bluebottle/funding/tests/test_admin.py
```diff
@@ -153,6 +154,18 @@ def test_donation_reward(self):
...
+        reported_time = f'{defaultfilters.date(curr_time)}, {defaultfilters.time(curr_time)}'
+        self.assertTrue(reported_time in response.content.decode('utf-8'))
```
gannetson, line 167: Better to explicitly check that this value appears in the first column of the table.

gannetson (changes requested):

bluebottle/funding/tests/test_admin.py
gannetson: Make sure to do the `self.client.get` after you set that value.
Also, you might want to use `self.app.get` from the webtest stack. You probably have to change the `TestCase` this extends or create a new `TestCase`. Webtest (the one you used in your other tests) has html searching features, which you can use to find the first column of the displayed table.

bluebottle/funding/tests/test_admin.py
```diff
@@ -153,6 +154,18 @@ def test_donation_reward(self):
...
         self.assertFalse(third.title in response.content.decode('utf-8'))
 
+    def test_contribution_date(self):
+        donation = DonorFactory.create(
+            activity=self.funding,
+            amount=Money(100, 'EUR')
+        )
+        donation.states.succeed(save=True)
+        self.client.force_login(self.superuser)
+        response = self.client.get(self.admin_url)
```
gannetson, line 164: `self.admin_url` refers to a list of Funding activities. You are trying to prove that the first column of a Donor list holds a value, right? So you should use `reverse('funding_donor_changelist')`. Best move this to another TestCase.

PR #5028, opened 2022-03-28
Filter initiatives by location

gannetson:


eodolphi (approved):

eodolphi:
Wouldn't it be easier to change this in the frontend? Would also save us from reindexing.

PR #5143, opened 2022-06-28
Hotfix/team serializer stuff

gannetson:


bluebottle/fsm/triggers.py
```diff
@@ -185,6 +186,8 @@ def _check_model_changed_triggers(self):
                         self._triggers.append(BoundTrigger(self, trigger))
 
     def execute_triggers(self, effects=None, **options):
+        if 'user' not in options and get_current_user():
+            options['user'] = get_current_user()
```
gannetson, line 190: Here's the nasty bit. @eodolphi do you think you have a better solution for this?

PR #5212, opened 2022-08-22
Feature: Filter homepage stats by year

gannetson:


bluebottle/statistics/renderers.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class StatisticsRenderer(JSONRenderer):
+
+    @classmethod
+    def build_json_resource_obj(cls, *args, **kwargs):
+        obj = super(StatisticsRenderer, cls).build_json_resource_obj(*args, **kwargs)
+        req = get_current_request()
+        if 'year' in req.GET:
+            obj['id'] = obj['id'] + '-' + req.GET['year']
```
gannetson, line 12: @eodolphi I think it would be better to set this in the serializer, but since the `id` is set in the renderer, this seems the only way...

PR #5214, opened 2022-08-23
Hotfix/translated lists

gannetson:


bluebottle/utils/views.py
```diff
@@ -232,9 +232,9 @@ def get_queryset(self):
 
 class TranslatedApiViewMixin(object):
     def get_queryset(self):
-        qs = super(TranslatedApiViewMixin, self).get_queryset().translated(
-            get_language()
-        )
+        qs = super(TranslatedApiViewMixin, self).get_queryset().active_translations(
+            get_current_language()
+        ).distinct('id').order_by('id')
```
eodolphi, line 237: Doesn't this mess up the ordering in dropdown menu's?

gannetson: I think doing a `distinct('translations__name')`, like we did in multiple places one of the causes of the mess up. As far as I can see all admin lists, filters and filters on front-end are good now.

eodolphi (approved):

PR #5217, opened 2022-08-29
Add link and external field to step blocks. This makes it possible to

eodolphi:
add a link to a step on the homepage.


bluebottle/utils/validators.py
```diff
@@ -108,6 +108,7 @@ def __init__(self, allowed_mimetypes=None, message=None, code=None):
 
     def __call__(self, value):
         try:
+            value.file.seek(0)
```
gannetson, line 111: What does this do?

gannetson (approved):

bluebottle/utils/validators.py
eodolphi: i had an issue where I could not upload a file because for some reason the file had already been read before this check. Possibly because the check was performed twice or something. This makes sure that the file is always read from the beginning and the mme type is properly returned.

PR #5249, opened 2022-09-20
Ticket/bb 20602 add fiscal year to user stats

gannetson:


eodolphi:
This misses the fiscal year for the periodic messages. Are we going to do that in a separate ticket?

eodolphi (approved):

bluebottle/activities/tasks.py
```diff
@@ -130,17 +131,17 @@ def recommend():
...
             settings = MemberPlatformSettings.objects.get()
             if settings.do_good_hours:
-                q1 = date.today().replace(month=1, day=1)
-                q2 = date.today().replace(month=4, day=1)
-                q3 = date.today().replace(month=7, day=1)
-                q4 = date.today().replace(month=10, day=1)
-                notification = None
+                offset = settings.fiscal_month_offset
                 today = date.today()
+                q1 = (date.today().replace(month=1, day=1) + relativedelta(months=offset)).replace(today.year)
```
eodolphi, line 140: I find this really hard to read. Why do `date.today().replace(month=1, day=1)` when `date(2000, 1, 1)` would also work.
I also find using `date.replace(<int>)` really unintuitive. Better use `date.replace(year=<int>)`

bluebottle/members/models.py
```diff
@@ -185,6 +187,20 @@ class MemberPlatformSettings(BasePlatformSettings):
         )
     )
 
+    @property
+    def fiscal_year_start(self):
+        offset = self.fiscal_month_offset
+        month_start = (datetime(2000, 1, 1) + relativedelta(months=offset)).month
+        return (now() + relativedelta(months=offset)).replace(month=month_start, day=1, hour=0, second=0)
```
eodolphi, line 194: This looks wrong. You take now, add the offset in months and then replace the month with the calculated month offset. Why not take the start of the year and set the month. 
I think this probably works, but it is not clear to me at all why.

eodolphi (changes requested):

PR #5323, opened 2022-12-12
Ticket/bb 21072 anonymize contributors

gannetson:


eodolphi (approved):
These needs quite a few more tests, but looks good  to me.

PR #5397, opened 2023-03-07
Disallow homepage gets on closed sites and for partner without permis…

eodolphi:
…sions


bluebottle/test/utils.py
```diff
@@ -586,7 +587,7 @@ def assertRequired(self, field):
         ]
         self.assertIn(field, error_fields)
 
-    @property
+    @ property
```
gannetson, line 590: Hmm... Is this new code style?

gannetson (approved):

PR #5398, opened 2023-03-08
Add footer banner to config

eodolphi:


Pieter-Rees (approved):

gannetson:
Maybe a test that it returns a url?

PR #5406, opened 2023-03-14
Remove quotes app completely

eodolphi:


gannetson:
Maybe it would be nice to clarn-up db too when removing this? It would require two steps though, but it is what we have done with a bunch of other obsolete apps, I think.

PR #5455, opened 2023-05-15
Upgrade packages. Add metaclass to activity segment admin form. Small…

eodolphi:
… fixes for upgrade


gannetson:
This is already in `release/pentest`

PR #5463, opened 2023-05-26
Hotfix/segment filter

gannetson:


eodolphi (changes requested):

bluebottle/activities/tests/test_api.py
```diff
@@ -626,8 +626,8 @@ def test_filter_segment(self):
         self.assertFacets(
             f'segment.{segment_type.slug}',
             {
-                f'segments.{matching_segment.pk}': len(matching),
-                f'segments.{other_segment.pk}': len(other)
+                f'{matching_segment.pk}': len(matching),
+                f'{other_segment.pk}': len(other)
```
gannetson, line 630: Still has to be cast to string though....
`f'{other_segment.pk}'` or `str(other_segment.pk)`
same difference :man_shrugging:

PR #5466, opened 2023-05-31
Change how we sort by date

gannetson:


gannetson:
Sorting now shows period activities first, sorted on end date, then slots by start date.
Only thing is that period activities without start- *and* enddate still end up at the back....
the `"missing" = "_first"` doesn't work as you would expect and I could'n't get  sorting by a `_script` to work either :(
Maybe get this in first and then yoyu acn give it another go @eodolphi ?

PR #5479, opened 2023-06-09
Fix ordering and filtering on succeeded activities

eodolphi:


bluebottle/activities/filters.py
```diff
@@ -168,35 +168,48 @@ def sort(self, search):
...
-
-            search = search.sort({
-                "dates.end": {
-                    "order": "asc",
-                    "nested": {
-                        "path": "dates",
-                        "filter": (
-                                Range(**{'dates.start': {'lte': start}}) &
+        elif 'upcoming' in self.filter_values:
+            if self.filter_values['upcoming'] == 'true':
```
gannetson, line 172: `if self.filter_values['upcoming'][0] == 'true':`

gannetson (changes requested):
One of the ifs is not correct yet, I think

PR #5591, opened 2023-09-21
Add action and link to plain image text block. Make text html, with l…

eodolphi:
…imited options


eodolphi:
Looks like this breaks some things. Some types changed and the way we use the components does not match the component anymore. This will need some testing and fixing of the types.

PR #5603, opened 2023-09-27
Return all participants when user is admin

eodolphi:


bluebottle/activities/views.py
```diff
@@ -373,12 +373,14 @@ def get_serializer_context(self, **kwargs):
 
     def get_queryset(self):
         if self.request.user.is_authenticated:
-            queryset = self.queryset.filter(
-                Q(user=self.request.user) |
-                Q(activity__owner=self.request.user) |
-                Q(activity__initiative__activity_manager=self.request.user) |
-                Q(status__in=('accepted', 'succeeded',))
-            ).order_by('-id')
+            if self.request.user.is_superuser:
```
gannetson, line 376: Maybe also `self.request.user.is_staff` ?

PR #5617, opened 2023-10-04
Change start date on homepage card

gannetson:


bluebottle/activities/serializers.py
```diff
@@ -106,12 +106,18 @@ class ActivityPreviewSerializer(ModelSerializer):
     collect_type = serializers.SerializerMethodField()
 
     def get_start(self, obj):
+        homepage = self.context['request'].GET.get('page[size]') == '4'
```
eodolphi, line 109: This looks really iffy to me. We identify that the actiivty is on the homepage by the page size? That will break some day.

eodolphi: I have a similar ticket on my name. When should we ever return non-upcoming slots on the card? I do not think we use that at all at the moment.

eodolphi (changes requested):

bluebottle/activities/serializers.py
gannetson: When a activity is succeeded, but shown on the homepage, we should show the last slot start date, I think.
I can change it to send an extra param, to indicate that it's on the homepage, or we could change the query to look if `filter[upcoming]` is not set. Maybe that's best

PR #5661, opened 2023-11-13
Ticket/bb 23223 add event to email

gannetson:


bluebottle/notifications/messages.py
```diff
@@ -98,6 +101,46 @@ def get_content_html(self, recipient):
     def get_content_text(self, recipient):
         return to_text.handle(self.get_content_html(recipient))
 
+    def get_event_data(self, recipient):
+        return self.event_data
```
eodolphi, line 105: Why do we need a getter for this?

eodolphi: Ah I see, sometimes you need the recipient. That does not seem like an ideal API. Why not always make it a method?

gannetson: I started with a property first of course. Though this was fine because I think a pattern with `get_prop_x` and `prop_x` happens in more places in Django (e.g. admin `get_fieldsets` and `fieldsets`), but fine to just use `get_prop_x` since there is no real benefit of also having the property.

gannetson: Fixed

PR #5736, opened 2024-01-15
Hide slot title when rendering an email that has a single related slot

eodolphi:


gannetson (approved):

gannetson:
It should only limit on the total slot-count of the activity. THis is limiting on how many slots you signed up for. That's why so many tests are failing.

gannetson (changes requested):
Change this so it only hide slot title if the `Activity` has multiple slots

PR #5768, opened 2024-02-26
Ticket/periodic activity

eodolphi:


bluebottle/collect/triggers.py
```diff
@@ -246,12 +225,6 @@ class CollectContributorTriggers(ContributorTriggers):
                     ParticipantAddedNotification,
                     conditions=[is_not_user]
                 ),
-                NotificationEffect(
-                    TeamMemberAddedMessage,
-                    conditions=[
-                        is_team_activity
-                    ]
-                ),
```
gannetson, line 254: Huh?! wat heeft dat ooit gedaan hier?

eodolphi: Ja geen idee.

bluebottle/time_based/models.py
```diff
@@ -684,14 +684,14 @@ def event_data(self):
...
+    (False, 'No, enter a location')
+)
 
-    ONLINE_CHOICES = (
-        (None, 'Not set yet'),
-        (True, 'Yes, participants can join from anywhere or online'),
-        (False, 'No, enter a location')
-    )
 
+class BaseActivity(TimeBasedActivity):
```
gannetson, line 694: This seems odd... to have `BaseActivity` extending `TimeBasedActivity`... Couldn't these additions just be part of `TImeBasedActivity`? Or name this something that describes what you're adding?

bluebottle/time_based/models.py
```diff
@@ -1066,6 +1129,38 @@ class JSONAPIMeta(object):
         resource_name = 'contributors/time-based/deadline-participants'
 
 
+class PeriodicSlot(TriggerMixin, models.Model):
```
gannetson, line 1132: We don't want to make slots polymorphi? Is there too little in common between the different slot types?

bluebottle/time_based/states/states.py
```diff
@@ -251,6 +222,72 @@ def can_succeed(self):
...
+
+@register(PeriodicActivity)
+class PeriodicActivityStateMachine(BaseTimeBasedStateMachine):
+    pass
+
+
+@register(PeriodActivity)
+class PeriodActivityStateMachine(ModelStateMachine):
+    pass
+
```
gannetson, line 238: Remove this

gannetson (approved):

bluebottle/time_based/models.py
eodolphi: I do not really know what to do here. Maybe RegistrationActivity is the right word.

bluebottle/time_based/models.py
eodolphi: I do not want to change the date slot model now. If we move date activities to registrations, then I'd make them a subclass.
And the to-be-decided activity will share a parent with this model.

bluebottle/time_based/states/states.py
```diff
@@ -157,37 +158,7 @@ class DateStateMachine(TimeBasedStateMachine):
...
-        description=_(
-            "The date of the activity has been changed to a date in the future. "
-            "The status of the activity will be recalculated."
-        ),
-    )
-
-
-@register(DeadlineActivity)
-class DeadlineActivityStateMachine(TimeBasedStateMachine):
+class BaseTimeBasedStateMachine(TimeBasedStateMachine):
```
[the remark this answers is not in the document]
eodolphi: That will break things for date activities. I will call this RegistrationActivityStateMachine.

PR #5915, opened 2024-06-12
Skip formsets when duplicating, so that we do not create wallposts or…

eodolphi:
… participats


bluebottle/activities/admin.py
```diff
@@ -286,6 +291,25 @@ class ActivityChildAdmin(PolymorphicChildModelAdmin, StateMachineAdmin):
     inlines = (FollowAdminInline, WallpostInline, )
     form = ActivityForm
 
+    skip_on_duplicate = [Contributor, Wallpost, Follow, Message, Update]
+
+    def get_formsets_with_inlines(self, request, obj=None):
+        formsets = super().get_formsets_with_inlines(request, obj)
+
+        if "_saveasnew" in request.POST:
+            __import__("ipdb").set_trace()
```
gannetson, line 300: Nope

gannetson (changes requested):

PR #5965, opened 2024-07-04
Don't generate upcoming slots for recurring act.

gannetson:


bluebottle/time_based/migrations/0101_auto_20240304_1439.py
```diff
@@ -67,7 +68,7 @@ def migrate_periodic_participants(apps, schema_editor):
 
         if activity.period:
 
-            while start < deadline:
+            while start < now():
```
eodolphi, line 71: Are we sure this is correct? Won't this generate loads of slots for old activities that already ended?

gannetson: a crap... it has to be whatever comes first...
The problem was that it generated upcoming slots for all activities with deadlines in the future
I'll adapt

eodolphi (approved):

PR #5966, opened 2024-07-05
Ticket/bb 24792 edit slot as admin

gannetson:


bluebottle/time_based/states/participants.py
```diff
@@ -494,3 +494,9 @@ class TeamScheduleParticipantStateMachine(ScheduleParticipantStateMachine):
 @register(PeriodicParticipant)
 class PeriodicParticipantStateMachine(RegistrationParticipantStateMachine):
     pass
+
+
+@register(PeriodParticipant)
```
eodolphi, line 499: This should not be nessecary. This is only needed when something went wrong in the migration. By adding these line, we will hide the actual error for ourselves.

gannetson: deloitte-uk tenant was not working for me.
I'll remove this

PR #6039, opened 2024-08-15
Release: Globalisation II

gannetson:


bluebottle/offices/admin.py
```diff
@@ -99,26 +102,50 @@ def activities(self, obj):
     fields = ('name', 'description', 'subregions_link', 'offices', 'activities')
 
 
-class RegionManagerAdminMixin:
+def region_manager_filter(queryset, user):
+
+    model = queryset.model
+    if model == Member and user.is_superuser:
```
eodolphi, line 108: Shouldn't superusers always see all models?

eodolphi: And adding this filter to the member admin would result in an error for region managers. So this does not really make sense.

eodolphi (changes requested):

PR #6180, opened 2024-11-28
Ticket/fix reaccept

gannetson:


eodolphi (approved):

eodolphi:
Maybe a script to fix the current contributions?

gannetson:
I think this would already be repaired with the incorrect contributions script we run weekly.

PR #6324, opened 2025-03-11
Script to fix rogue filters

gannetson:


scripts/fix_filters.py
```diff
@@ -0,0 +1,12 @@
...
+from bluebottle.initiatives.models import ActivitySearchFilter
+
+
+def run(*args):
+    for tenant in Client.objects.all().all():
+        with LocalTenant(tenant):
+            print(f'Fixing {tenant.name}')
+            ActivitySearchFilter.objects.filter(
+                type__in=['managing', 'status']
+            ).all().delete()
```
eodolphi, line 12: Shouldn't we just figure out why we create these filters?

eodolphi (changes requested):

scripts/fix_filters.py
gannetson: I tried to find out what might have created these filters. Couldn't find it yet.
This a script to quickly to remove it from all tenants when it pops up again. Maybe it doesn't have  to be ik the repo though...

PR #6341, opened 2025-03-18
Don't save filters on the fly

gannetson:


bluebottle/activities/filters.py
```diff
@@ -334,174 +315,183 @@ class ActivitySearch(Search):
...
+                            },
                         }
                     }
-                })
+                )
                 return search
 
         return search
 
     def __new__(cls, *args, **kwargs):
```
gannetson, line 476: This method has the only one relevant change

PR #6342, opened 2025-03-18
Fix initiative/image/theme requirement

gannetson:
image


bluebottle/activities/states.py
```diff
@@ -77,11 +82,16 @@ def is_valid(self):
 
     def initiative_is_approved(self):
```
gannetson, line 83: changes here

eodolphi (approved):

PR #6351, opened 2025-03-21
Fix wallposts

gannetson:


bluebottle/updates/serializers.py
```diff
@@ -124,7 +126,7 @@ class JSONAPIMeta(object):
     'small': '150x150',
     'medium': '800x450',
     'large': '1600x900',
-    'full': '1600',
+    'full': '1600x1600',
 }
```
eodolphi, line 130: I purposely made this `1000` to not crop that image at all.

PR #6355, opened 2025-03-24
Do not use autocomplete fields. This already works by default and bre…

eodolphi:
…aks because of duplicate jquery imports


bluebottle/initiatives/admin.py
```diff
@@ -212,8 +212,6 @@ def valid(self, obj):
             format_html_join("", "<li>{}</li>", ((value,) for value in errors))
         )
 
-    autocomplete_fields = ['activity_managers']
```
gannetson, line 215: this will then show a long list with all users where you can select multiple, right? If you have a lot of users this will make these pages really slow.

PR #6393, opened 2025-04-14
Allow staff users to edit activities

eodolphi:


bluebottle/initiatives/permissions.py
```diff
@@ -20,5 +20,7 @@ class InitiativeOwnerPermission(ResourceOwnerPermission):
     def has_object_action_permission(self, action, user, obj):
         return (
             super().has_object_action_permission(action, user, obj) or
-            user in obj.activity_managers.all()
+            user in obj.activity_managers.all() or
+            user.is_staff
```
gannetson, line 24: Best add `user.is_superuser` too

gannetson (changes requested):

PR #6606, opened 2025-09-02
Hour regdsitration

gannetson:


bluebottle/utils/email_backend.py
```diff
@@ -84,6 +84,52 @@ def _send(self, email_message):
 DKIMBackend = TenantAwareBackend
 
 
+class PlainBackend(EmailBackend):
```
eodolphi, line 87: What is this used for?

PR #6969, opened 2026-05-18
Fix linked images

gannetson:


eodolphi (approved):

bluebottle/activity_links/models.py
```diff
@@ -4,7 +4,6 @@
 from djmoney.money import Money
 from polymorphic.models import PolymorphicModel, PolymorphicManager
 
-from bluebottle.activity_pub.models import Follow, Create
```
eodolphi, line 7: Are you sure these should be removed?

eodolphi (changes requested):

PR #6976, opened 2026-05-21
Fix celery task imports, retry create_task so that Start activities t…

eodolphi:
…hat are early will still transition the link


gannetson (approved):

bluebottle/activity_pub/views.py
```diff
@@ -44,7 +44,10 @@ class OrganizationView(ActivityPubView):
     queryset = Organization.objects.all()
 
 
-@shared_task()
+@app.task(
+    autoretry_for=(Exception,), retry_backoff=True, retry_kwargs={'max_retries': 5},
+    name="bluebottle.activity_pub.adapters.publish_activities"
```
gannetson: rename this?

PR #7023, opened 2026-06-17
Fix when KYC mails get triggered

gannetson:


eodolphi (approved):

bluebottle/funding_stripe/triggers.py
```diff
@@ -198,23 +201,23 @@ def has_new_requirements(self):
             effects=[
                 NotificationEffect(
                     GrantApplicationPayoutAccountMarkedIncomplete,
-                    conditions=[has_grant_application]
+                    conditions=[has_granted_grant_application]
                 ),
                 NotificationEffect(
                     FundingPayoutAccountMarkedIncomplete,
-                    conditions=[is_not_public, has_funding_campaign],
+                    conditions=[is_not_public, has_active_funding_campaign],
```
eodolphi: These conditions are now the save as the 'live payout account'. That can't be right. Now always when you receive the 'live payout campaign' message you also receive this. 
This message is also meant for users that are setting up their campaign. Draft campaigns that submit their kyc account should receive this message when the account goes from submitted to incomplete.

eodolphi (changes requested):

eodolphi (approved):

PR #7036, opened 2026-06-30
Ticket/targeted posts

gannetson:


bluebottle/updates/views.py
```diff
@@ -73,10 +85,49 @@ class ActivityUpdateList(JsonApiViewMixin, ListAPIView):
...
-            activity_id=self.kwargs['activity_pk']
-        ).filter(parent__isnull=True)
+        queryset = self.get_visible_queryset()
+        audience_filter = self.request.query_params.get('filter[audience]')
+        if audience_filter in (AudienceChoices.everyone, AudienceChoices.contributors):
+            queryset = queryset.filter(audience=audience_filter)
+        return queryset
+
+    def list(self, request, *args, **kwargs):
+        visible_queryset = self.get_visible_queryset()
```
eodolphi: It is hard to tell why this is overwritten, but setting the `audience` meta field should not happen here. There serializer should do that.

See here: https://django-rest-framework-json-api.readthedocs.io/en/stable/usage.html#meta on how to do that.

eodolphi (changes requested):

PR #7087, opened 2026-07-24
use activity title rather than content_object title

WolfHakase:


gannetson (approved):

gannetson:
Ah yes. In the past we also had update wall on Initiative page. We ditched that one.

PR #7096, opened 2026-07-29
Add team activities to AP

gannetson:


bluebottle/activity_pub/serializers/federated_activities.py
WolfHakase, line 1: I feel like we should be able to split up this file without introducing circular dependencies. I file with nearly 1500 lines feels hefty.

bluebottle/time_based/effects/teams.py
WolfHakase: I may not understand exactly what is happening here, but with a remote captain user would be None, right? So would this match any registration with user=None on that activity? Or would it simply miss the correct remote_user match? It seems the create path does set remote_user, but the lookup may not. Could this potentially cause a team being attached to the wrong registration?

bluebottle/activity_pub/serializers/federated_activities.py
```diff
@@ -1005,3 +1046,401 @@ def create(self, validated_data):
...
+        if validated_data.get('remote_user'):
+            filters['remote_user'] = validated_data['remote_user']
+        if validated_data.get('user'):
+            filters['user'] = validated_data['user']
+
+        registration = TeamScheduleRegistration.objects.filter(**filters).first()
+        member_filters = {
+            k: v for k, v in filters.items() if k != 'activity'
+        }
+        team_member = registration.teams.first().team_members.filter(
```
WolfHakase: Could registration technically be None here? Would that cause a crash? Do we need a guard against this situation?

bluebottle/activity_pub/effects.py
```diff
@@ -316,6 +316,98 @@ def __str__(self):
...
+    def post_save(self, **kwargs):
+        registration = self.instance.registration
+        if registration:
+            adapter.sync(registration)
+
+    def __str__(self):
+        return str(_('Notify source platform of team join'))
+
+
+class SendAddToTeamEffect(Effect):
```
WolfHakase, line 355: This seems to run on initiate and reapply, but not on readd and accept. Inbound add seems to restore removed/rejected members. withdraw then reapply seems to sync. But remove then readd may not notify the supplier. Am I missing something?

Initially I read readd as "read" with two d's and thought it was a typo. I feel like the name would be more clear as re_add or reAdd, or something. But, I understand the choice.

bluebottle/activity_pub/serializers/federated_activities.py
```diff
@@ -1005,3 +1046,401 @@ def create(self, validated_data):
...
+            user=user,
+        ).first()
+        if existing:
+            team = existing.teams.first()
+            try:
+                if team and team.status == 'withdrawn':
+                    team.states.rejoin(save=True)
+                elif existing.status == 'withdrawn':
+                    existing.states.restore(save=True)
+            except TransitionNotPossible:
```
WolfHakase, line 1136: Is this correct? Would this result in the caller treating a failure as a success?

bluebottle/time_based/models.py
```diff
@@ -1730,7 +1729,18 @@ class Team(TriggerMixin, models.Model):
         'members.Member',
         verbose_name=_('Team captain'),
         related_name='team_captains',
-        on_delete=models.CASCADE
+        on_delete=models.CASCADE,
```
WolfHakase, line 1732: When we delete a remote captain do we want to delete it's entire team? The field is nullable.

WolfHakase (commented):
On top of the comments below, I also think the test suite could be stronger. But because this is all very much in flux I understand the choice.

bluebottle/activity_pub/serializers/federated_activities.py
gannetson: We can split the serializers up a bit more after we merge this into the demo branch.

bluebottle/activity_pub/serializers/__init__.py
```diff
@@ -81,6 +83,12 @@ def __init__(self, *args, **kwargs):
         self.resource_type_model_mapping['ScheduleSlot'] = ScheduleSlot
         self.resource_type_model_mapping['PeriodicSlot'] = PeriodicSlot
         self.resource_type_model_mapping['DateActivitySlot'] = DateActivitySlot
+        # Team schedule registrations serialize as Join outbound, but inbound Join
+        # adopt must stay on Contributor (polymorphic dispatch handles teams).
```
eodolphi, line 87: I do not understand what that means. And how that explains the line below.

bluebottle/activity_pub/serializers/federated_activities.py
```diff
@@ -1005,3 +1046,530 @@ def create(self, validated_data):
...
+    class Meta:
+        model = LocalTeam
+        fields = FederatedObjectBaseSerializer.Meta.fields + (
+            'name', 'summary', 'attributed_to', 'captain',
+        )
+
+    def create(self, validated_data):
+        validated_data.pop('id', None)
+        captain = validated_data.pop('user', None)
+        activity = validated_data['activity']
```
eodolphi, line 1069: Why take the activity here? We do not seem to use it.

bluebottle/activity_pub/serializers/federated_activities.py
```diff
@@ -1005,3 +1046,530 @@ def create(self, validated_data):
...
+                else:
+                    remote_user = field.save()
+
+        team = LocalTeam.objects.create(
+            activity=activity,
+            user=user,
+            remote_user=remote_user,
+            name=validated_data.get('name'),
+            description=validated_data.get('description'),
+            registration=validated_data.get('registration'),
```
eodolphi, line 1092: passing in `**validated_data` would make this quite a bit more robust against future changes.

bluebottle/activity_pub/serializers/federated_activities.py
```diff
@@ -1005,3 +1046,530 @@ def create(self, validated_data):
...
+        )
+        return team
+
+
+class TeamInstrumentField(serializers.Field):
+    """Outbound nested Team for a team-schedule Join."""
```
eodolphi, line 1098: Why do we need a different type of serializer here? Why does the normal FederatedTeamSerializer not work?

bluebottle/activity_pub/serializers/federated_activities.py
```diff
@@ -1005,3 +1046,530 @@ def create(self, validated_data):
...
+        if instrument_iri:
+            ap_team = ActivityPubModel.objects.from_iri(instrument_iri)
+            if ap_team and not ap_team.adopted:
+                ap_team.adopted = team
+                if (
+                    isinstance(ap_team, ActivityPubTeam) and
+                    not ap_team.attributed_to_id and
+                    hasattr(activity, 'activity_pub_model')
+                ):
+                    ap_team.attributed_to = activity.activity_pub_model
```
eodolphi, line 1179: What is  `.attributed_to` here, and how does it differ from `ap_team.adopted` or `ap_team.origin`?

bluebottle/activity_pub/serializers/federated_activities.py
```diff
@@ -1005,3 +1046,530 @@ def create(self, validated_data):
...
+
+    def get_target(self, obj):
+        if hasattr(obj.team, 'activity_pub_model'):
+            return obj.team.activity_pub_model.pub_url
+        if hasattr(obj.team, 'origin'):
+            return obj.team.origin.pub_url
+
+    def to_representation(self, instance):
+        data = super().to_representation(instance)
+        target = self.get_target(instance)
```
eodolphi, line 1289: Why wouldn't this be set by default? The target field should just be set by the field.

bluebottle/activity_pub/serializers/federated_activities.py
```diff
@@ -1005,3 +1046,530 @@ def create(self, validated_data):
...
+        if target:
+            data['target'] = target
+        return data
+
+    @staticmethod
+    def _resource_iri(value):
```
eodolphi, line 1295: This should not be a private class helper. This should be an util function or something.

bluebottle/activity_pub/serializers/federated_activities.py
```diff
@@ -1005,3 +1046,530 @@ def create(self, validated_data):
...
+            team=team, user=None, remote_user=remote_user
+        ).first()
+        if existing:
+            try:
+                if existing.status == 'withdrawn':
+                    existing.states.reapply(save=True)
+                elif existing.status == 'removed':
+                    existing.states.readd(save=True)
+                elif existing.status == 'rejected':
+                    existing.states.accept(save=True)
```
eodolphi, line 1439: These transitions should not happen in the serializer.

bluebottle/activity_pub/serializers/federated_activities.py
```diff
@@ -1005,3 +1046,530 @@ def create(self, validated_data):
...
+            except TransitionNotPossible:
+                pass
+            return existing
+
+        member = TeamMember(
+            team=team,
+            user=None,
+            remote_user=remote_user,
+        )
+        member.execute_triggers()
```
eodolphi, line 1449: Why do we execute triggers separately from saving. Saving also executes the triggers by default.

bluebottle/activity_pub/serializers/federated_activities.py
```diff
@@ -1005,3 +1046,530 @@ def create(self, validated_data):
...
+
+        if not self._platform_may_modify_event(platform, event):
+            raise serializers.ValidationError({
+                'target': 'Platform is not authorized to add members to this team',
+            })
+
+        return platform
+
+    def create(self, validated_data):
+        validated_data.pop('id', None)
```
eodolphi, line 1369: A lot of these save method are becoming really complicated. And the should not be. We should push the complication to the fields, and simplify this.

bluebottle/activity_pub/serializers/federated_activities.py
```diff
@@ -1005,3 +1046,530 @@ def create(self, validated_data):
...
+    team = RelatedTeamSlotField()
+
+    class Meta(ScheduleSlotsSerializer.Meta):
+        model = TeamScheduleSlot
+        fields = ScheduleSlotsSerializer.Meta.fields + ('team',)
+
+    def _prepare_slot_validated_data(self, validated_data):
+        """
+        Drop inbound status (let ModelChangedTriggers run schedule) and hydrate
+        nested federated objects such as location.
```
eodolphi, line 1508: Why does this serializer need to do this, and other do not? Why would validated data ever have a `status` value? There is no field named status.

bluebottle/activity_pub/serializers/federated_activities.py
```diff
@@ -1005,3 +1046,530 @@ def create(self, validated_data):
...
+        if origin and hasattr(origin, 'adopted'):
+            origin.adopted = result
+            origin.save()
+
+    def create(self, validated_data):
+        iri = validated_data.pop('id', None)
+        validated_data = self._prepare_slot_validated_data(validated_data)
+        existing = self._team_slot_to_reuse(validated_data)
+
+        if existing:
```
eodolphi, line 1561: We should not call update in the create method. Instead overwrite the save method so that `self.instance` gets set.

bluebottle/activity_pub/serializers/federated_activities.py
```diff
@@ -1005,3 +1046,530 @@ def create(self, validated_data):
...
+            result = serializers.ModelSerializer.update(self, existing, validated_data)
+        else:
+            result = serializers.ModelSerializer.create(self, validated_data)
```
eodolphi, line 1564: We should use `super` here, since this might skip crucial actions in the `create` method of the parent class.

bluebottle/activity_pub/serializers/json_ld.py
```diff
@@ -250,6 +252,21 @@ def to_internal_value(self, data):
             return event
 
 
+class RelatedTeamField(RelatedField):
```
eodolphi, line 255: This does not look different from the normal related serializer? Why is this specific to teams?

bluebottle/activity_pub/tests/test_team_add_authorization.py
```diff
@@ -0,0 +1,139 @@
...
+from bluebottle.activity_pub.serializers.federated_activities import (
+    TeamMemberAddSerializer, TeamScheduleSlotsSerializer,
+)
+from bluebottle.activity_pub.tests.factories import OrganizationFactory, PersonFactory
+from bluebottle.cms.models import SitePlatformSettings
+from bluebottle.test.utils import BluebottleTestCase
+from bluebottle.time_based.tests.factories import ScheduleActivityFactory, TeamFactory
+
+
+class TeamMemberAddAuthorizationTestCase(BluebottleTestCase):
```
eodolphi, line 17: Why do we have a specific test case for team authorization? What is different about authorization with teams then without? Can't this be tested in a `test_authorization` file?

Besides, this does not look to be about authorization but about validation. If it is about authorization, the logic should not live in the serializers and the view should have a permission class to verify this.

bluebottle/activity_pub/models.py
```diff
@@ -576,11 +590,58 @@ def activity_type(self):
...
 
 
+class Team(ActivityPubModel):
+    """
+    Federated team for team-schedule activities. Maps to time_based.Team.
+    """
+    name = models.CharField(max_length=300, null=True, blank=True)
+    summary = models.TextField(null=True, blank=True)
+
+    attributed_to = models.ForeignKey(
```
eodolphi, line 609: Why is this field called `attributed_to`? That should link to the actor that owns or created the original object the resource is a copy of. The activity is not the owner and the activity did not create this team.

bluebottle/activity_pub/models.py
```diff
@@ -915,6 +976,14 @@ class Join(Activity):
     """Sent by a follower when a user joins an Event"""
     object = models.ForeignKey(ActivityPubModel, on_delete=models.CASCADE)
     motivation = models.TextField(null=True, blank=True)
+    instrument = models.ForeignKey(
```
eodolphi, line 979: Why did we call this field `instrument`? I do not understand how that related to a team.

bluebottle/activity_pub/models.py
```diff
@@ -938,6 +1007,39 @@ def default_recipients(self):
...
+        Person,
+        on_delete=models.CASCADE,
+        related_name='added_by',
+    )
+    target = models.ForeignKey(
+        Team,
+        on_delete=models.CASCADE,
+        related_name='adds',
+    )
+    platform = models.ForeignKey(Organization, null=True, on_delete=models.CASCADE)
```
eodolphi, line 1022: Why does this resource have a specific platform relation? Can't we infer the platform from the activity the team belongs to, and if not why don't other `Activity` models have this. And who is the actor here? Isn't that the same as the platform?

More general, why do we use `Add` and not `Join` here? And why do we do this specific join with and subject, object and target. Instead of doing subject, object. Eg. Ernst joined this team.

bluebottle/time_based/effects/slots.py
```diff
@@ -9,11 +9,16 @@ class CreateTeamSlotParticipantsEffect(Effect):
...
     def post_save(self, **kwargs):
-        for team_member in self.instance.team.team_members.filter(status__in=['new', 'accepted']).all():
+        if not self.instance.team_id:
+            return
+        for team_member in self.instance.team.team_members.filter(status='active').all():
             slot = self.instance
             slot.participants.get_or_create(
                 user=team_member.user,
+                remote_user=team_member.remote_user,
+                team_member=team_member,
```
eodolphi, line 19: Why does a slot participant have a team_member field? Can't that be inferred?

bluebottle/time_based/triggers/participants.py
```diff
@@ -1145,6 +1145,21 @@ def team_is_accepted(effect):
...
             ],
         ),
+        TransitionTrigger(
+            TeamScheduleParticipantStateMachine.schedule,
+            effects=[
+                SendJoinSlotEffect,
+                SyncRelatedEvent,
+            ],
+        ),
+        ModelChangedTrigger(
```
eodolphi, line 1155: Does this ever happen? Isn't there always a slot created in a different effect?

PR #7101, opened 2026-07-31
Fix autocomplete so we don't load 9000 segments in

gannetson:
activity detail admin page

### DIFF
--- a/bluebottle/segments/widgets.py
+++ b/bluebottle/segments/widgets.py
@@ -1,0 +1,23 @@
+from django.contrib import admin
+from django.contrib.admin.widgets import AutocompleteSelect, AutocompleteSelectMultiple
+
+
+class SegmentTypeAutocompleteMixin:
+    def __init__(self, field, segment_type_id, admin_site=None, attrs=None):
+        self.segment_type_id = segment_type_id
+        super().__init__(field, admin_site or admin.site, attrs=attrs)
+
+    def build_attrs(self, base_attrs, extra_attrs=None):
+        attrs = super().build_attrs(base_attrs, extra_attrs)
+        url = attrs['data-ajax--url']
+        separator = '&' if '?' in url else '?'
+        attrs['data-ajax--url'] = f'{url}{separator}segment_type={self.segment_type_id}'
+        return attrs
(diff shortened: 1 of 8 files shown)


bluebottle/utils/monkey_patch_jet.py
```diff
@@ -48,5 +48,34 @@ def formfield_for_manytomany(self, db_field, request, **kwargs):
...
+    RelatedFieldListFilter / SortedRelatedFieldListFilter then load every related
+    object (and parler translations) on every change page. Filters are not needed
+    to resolve sibling objects, so skip them.
+    """
+    model_admin = admin_site._registry.get(model)
+    if model_admin is None:
+        return
+
+    original_get_list_filter = model_admin.get_list_filter
+    model_admin.get_list_filter = lambda request: []
```
WolfHakase: Is this thread-safe? If multiple threads would patch the get_list_filter method, could it happen that one restores the wrong method, due to timing issues?

bluebottle/segments/widgets.py
```diff
@@ -0,0 +1,23 @@
...
+        separator = '&' if '?' in url else '?'
+        attrs['data-ajax--url'] = f'{url}{separator}segment_type={self.segment_type_id}'
+        return attrs
+
+
+class SegmentAutocompleteSelectMultiple(SegmentTypeAutocompleteMixin, AutocompleteSelectMultiple):
+    pass
+
+
+class SegmentAutocompleteSelect(SegmentTypeAutocompleteMixin, AutocompleteSelect):
```
WolfHakase: This seems unused. Do we need it?

WolfHakase (approved):
Approved, with questions.

bluebottle/utils/monkey_patch_jet.py
gannetson: I removed the patch. It was trying to fix a related problem.

bluebottle/segments/widgets.py
gannetson: Right. Removed it. we can easily add it if we need it somewhere.

PR #7103, opened 2026-07-31
Change accept form

gannetson:


bluebottle/activity_pub/templates/admin/activity_pub/follow/accept_publish_mode.html
```diff
@@ -12,26 +12,24 @@
...
+                </div>
+                <div class="form-row">
+                    <label>{% trans 'Adoption type:' %}</label> {{ obj.short_adoption_type }}
+                </div>
+                <div class="form-row">
+                    {% trans 'Select how activities from this partner should be shared before accepting the follow request.' %}
+                </div>
+                <div class="form-row">
+                    <div>
+                        {{ form.as_table }}
```
WolfHakase: Does this work? I think Django form.as_table emits bare < tr > rows and expects a wrapping < table >. Did you mean form.as_p or form.as_div?

WolfHakase (approved):

bluebottle/activity_pub/templates/admin/activity_pub/follow/accept_publish_mode.html
gannetson: It didn't show tr/td tags for me... but i changed it to more explicit elements anyway

gannetson: > the browser will silently "fix" the invalid HTML because <tr> elements are only valid inside a table. That can make it look as though Django changed, when in fact it's the browser's HTML parser rearranging the DOM.

PR #7106, opened 2026-07-31
Only share upcoming ad previously published slots

gannetson:


bluebottle/activity_pub/tests/test_integration.py
```diff
@@ -1502,6 +1502,67 @@ def test_publish(self):
         with LocalTenant(self.other_tenant):
             self.assertEqual(self.event.sub_event.count(), 3)
 
+    def test_publish_skips_past_unpublished_slots(self):
+        from django.utils.timezone import now
```
WolfHakase: Is there a reason to have this local import twice, rather than having a file level import?

bluebottle/activity_pub/tests/test_integration.py
```diff
@@ -1502,6 +1502,67 @@ def test_publish(self):
...
+        self.assertFalse(past_slot.activity_pub_models.exists())
+        self.assertTrue(upcoming_slot.activity_pub_models.exists())
+        self.assertEqual(self.model.activity_pub_model.sub_event.count(), 1)
+        self.assertEqual(
+            self.model.activity_pub_model.sub_event.get().origin_id,
+            upcoming_slot.id,
+        )
+
+    def test_publish_keeps_already_published_past_slots(self):
+        from django.utils.timezone import now
```
WolfHakase: Is there a reason to have this local import twice, rather than having a file level import?

bluebottle/activity_pub/serializers/federated_activities.py
```diff
@@ -703,7 +703,7 @@ def get_contributor_count(self, obj):
         return obj.participants.filter(status__in=['accepted', 'new', 'succeeded']).count()
 
     def create(self, validated_data):
-        slots = validated_data.pop('slots', [])
+        slots = validated_data.pop('publishable_slots', validated_data.pop('slots', []))
```
WolfHakase: Python always evaluates the default, so this always pops slots even when publishable_slots is present. Should we instead:
```
slots = validated_data.pop('publishable_slots', None)
if slots is None:
    slots = validated_data.pop('slots', [])
```

It's more code, but more explicit and performs slightly better.

bluebottle/time_based/models.py
```diff
@@ -256,6 +256,20 @@ def start(self):
     def active_slots(self):
         return self.slots.filter(status__in=['open', 'full', 'running', 'finished']).order_by('start', 'id')
 
+    @property
+    def publishable_slots(self):
+        from django.db.models import Q, F, ExpressionWrapper, DateTimeField
```
WolfHakase: If possible, "prefer top-of-file" as defined in PEP8.

bluebottle/time_based/models.py
```diff
@@ -256,6 +256,20 @@ def start(self):
...
+
+        current = timezone.now()
+        return self.slots.annotate(
+            _end=ExpressionWrapper(
+                F('start') + F('duration'),
+                output_field=DateTimeField()
+            )
+        ).filter(
+            Q(_end__gte=current) | Q(start__gte=current) | Q(activity_pub_models__isnull=False)
```
WolfHakase: Does start_gte matter here? I think it may matter when duration and _end is null. But it's a bit unclear and is likely to accidentally be refactored away. Could we validate it's usage with extra test coverage?

bluebottle/time_based/models.py
```diff
@@ -513,6 +527,17 @@ def activity_pub_model(self):
         except ObjectDoesNotExist:
             raise AttributeError('activity_pub_model')
 
+    @property
+    def is_activity_pub_publishable(self):
```
WolfHakase, line 528: I think this property and the publishable_slots ORM queryset filter do essentially the same. That's fine. But what is keeping them the same? If we change one and keep the other as is they're out of sync.

WolfHakase (approved):

bluebottle/time_based/models.py
gannetson: fixed

bluebottle/time_based/models.py
gannetson: I just kept the `start__gte=current` It is silly to publish slots that are already started.

bluebottle/time_based/models.py
gannetson: Nice  catch. Will make sure they one uses the other.

bluebottle/activity_pub/serializers/federated_activities.py
gannetson: All good

PR #7108, opened 2026-07-31
Improve Remote contributor admin

gannetson:


bluebottle/activities/admin.py
```diff
@@ -92,11 +92,91 @@
...
+
+    status_label.short_description = _('Status')
+
+
+class RemoteMemberPlatformFilter(admin.SimpleListFilter):
+    title = _('Platform')
+    parameter_name = 'platform'
+
+    def lookups(self, request, model_admin):
+        from bluebottle.activity_pub.models import Organization
```
WolfHakase: If possible imports at the top. Unless, this creates a circular dependency.

bluebottle/settings/admin_dashboard.py
```diff
@@ -188,12 +188,15 @@
                 "url": "/en/admin/activity_pub/publishedactivity/",
                 "label": _("Shared activities"),
             },
+            {
+                "name": "activities.remotemember",
+                "enabled": "cms.SitePlatformSettings.is_publishing_activities",
+            },
             {
                 "name": "activity_pub.receivedactivity",
```
WolfHakase, line 196: Is this correct? Shouldn' t this be "remotecontributers"  or something?

bluebottle/time_based/admin.py
```diff
@@ -573,24 +573,44 @@ class BaseRegistrationAdminInline(TabularInlinePaginated):
...
-    def get_fields(self, request, obj=None):
-        fields = super().get_fields(request, obj)
-        try:
-            obj.activity_pub_model
-            return list(fields) + ['remote_user', 'platform']
-        except (Activity.activity_pub_model.RelatedObjectDoesNotExist, AttributeError):
-            pass
-
-        return fields
+    def participant(self, obj):
```
WolfHakase: There seems to be a nearly identical implementation in BaseContributorInline and BaseRegistrationAdminInline. We could extract that so the behaviour won't drift.

WolfHakase (approved):

bluebottle/settings/admin_dashboard.py
gannetson: The `label` was rong not `name`. Fixed

PR #7116, opened 2026-08-06
Cancel adpoted activities

gannetson:


bluebottle/activity_pub/effects.py
```diff
@@ -126,6 +126,66 @@ def __str__(self):
...
+            )
+
+    def _local_contributors(self):
+        exclude_statuses = ('withdrawn', 'removed', 'rejected')
+        if hasattr(self.instance, 'participants'):
+            return self.instance.participants.filter(
+                remote_user__isnull=True,
+            ).exclude(status__in=exclude_statuses)
+        if hasattr(self.instance, 'contributors'):
+            from bluebottle.activities.models import Organizer
```
WolfHakase: Is this inline import needed?

bluebottle/activity_pub/effects.py
```diff
@@ -126,6 +126,66 @@ def __str__(self):
         return str(_('Publish that the activity has been adopted'))
 
 
+class UnpublishAdoptionEffect(Effect):
+    """
+    Announce that an adopted activity or slot is no longer adopted.
+    """
+    display = True
+    template = 'admin/activity_pub/unpublish_adoption_effect.html'
```
WolfHakase, line 141: Missing template?

bluebottle/activity_links/triggers.py
```diff
@@ -17,4 +17,8 @@ class LinkedActivityTriggers(TriggerManager):
             LinkedActivityStateMachine.succeed,
             effects=[PublishAdoptionEffect]
         ),
+        TransitionTrigger(
+            LinkedActivityStateMachine.cancel,
```
WolfHakase, line 21: This transistion will AttributeError?

bluebottle/activity_pub/effects.py
```diff
@@ -126,6 +126,66 @@ def __str__(self):
...
+    def _leave_target(self, contributor):
+        if hasattr(contributor, 'slot') and contributor.slot and getattr(contributor.slot, 'origin', None):
+            return contributor.slot.origin
+        return getattr(self.instance, 'origin', None)
+
+    @property
+    def is_valid(self):
+        return (
+            getattr(self.instance, 'origin', False) or
+            isinstance(self.instance, LinkedActivity)
```
WolfHakase: LinkedActivity expost event, not origin. So cancel will AttributeError. This same pattern already exists on PublishAdoptionEffect (L113 / L121).

bluebottle/activity_pub/models.py
```diff
@@ -883,17 +914,53 @@ class Reject(Activity):
...
+            if create:
+                yield create.actor
 
     def save(self, *args, **kwargs):
         created = not self.pk
         super().save(*args, **kwargs)
 
-        if created and not self.is_local and isinstance(self.object, Join):
+        if not created or self.is_local:
+            return
```
WolfHakase, line 920: This early return means _unadopt() will only run when Reject is received. The adopter’s local Accept created by PublishAdoptionEffect is never removed. On restore + reapprove that can leave duplicate Accepts. Either delete local Accept in UnpublishAdoptionEffect.post_save, or also run unadopt for local Event/SubEvent Rejects.

bluebottle/activity_pub/effects.py
```diff
@@ -126,6 +126,66 @@ def __str__(self):
...
+    Announce that an adopted activity or slot is no longer adopted.
+    """
+    display = True
+    template = 'admin/activity_pub/unpublish_adoption_effect.html'
+
+    def post_save(self, **kwargs):
+        event = self.instance.origin
+        actor = get_platform_actor()
+
+        Reject.objects.create(actor=actor, object=event)
```
WolfHakase, line 158: Creating Reject and then Leave(forced=True) per contributor means the supplier ends the same remotes twice (Reject._unadopt + Leave.transition). If Reject is enough for federation, the Leave loop may be redundant noise; if both are required, a short comment on why would help.

bluebottle/activity_pub/models.py
```diff
@@ -883,17 +914,53 @@ class Reject(Activity):
...
+
+        if isinstance(self.object, Join):
             registration = Registration.objects.get(
                 user=self.object.actor.origin, activity=self.object.object.adopted
             )
             registration.states.reject(save=True)
+        elif isinstance(self.object, (Event, SubEvent)):
+            self._unadopt()
```
WolfHakase, line 928: _unadopt already fails/cancels all remotes from that actor. Overlaps with forced Leave handling in Leave.transition (e.g. L1174, L1189, L1215).

bluebottle/activity_pub/effects.py
```diff
@@ -126,6 +126,66 @@ def __str__(self):
...
+            if target is None:
+                continue
+            Leave.objects.create(
+                actor=contributor.user.activity_pub_model,
+                object=target,
+                forced=True,
+            )
+
+    def _local_contributors(self):
+        exclude_statuses = ('withdrawn', 'removed', 'rejected')
```
WolfHakase: Exclude list is withdrawn / removed / rejected, but cancel triggers often fail participants first (e.g. deeds L250, then UnpublishAdoptionEffect at L252). failed / cancelled contributors still get forced Leave. Either extend the exclude list or run unpublish before failing participants.

WolfHakase (changes requested):
Some of these comments are AI generated. So verify them before taking them to heart. But according to my agent there's some bugs. Please have a look. Let me know if it's mumbo-jumbo. Then I'll look at it again.

bluebottle/activity_pub/effects.py
gannetson: Correct. Removed the leave loop here. Participants will send leave individually through triggers.

bluebottle/activity_pub/models.py
gannetson: Current Reject.save already does the second option: _unadopt() runs for local and remote Event/SubEvent Rejects; the local skip is only for Join.

PR #7117, opened 2026-08-07
Clean-up remote participant admin

gannetson:


bluebottle/activities/admin.py
```diff
@@ -908,13 +959,12 @@ def get_detail_fields(self, request, obj):
                 'description_preview' if field == 'description' else field
                 for field in detail_fields
             )
-        if Location.objects.exists() and not settings.enable_office_restrictions:
-            detail_fields += ('office_location',)
-        if obj:
             detail_fields = tuple(
                 'display_image' if field == 'image' else field
                 for field in detail_fields
             )
```
WolfHakase, line 917: The deleteion of these lines makes the indentation of this block match the above if statement. Is that correct? I think this changed the behaviour to image and display image only swapping when the obj is adopted.

bluebottle/activities/admin.py
```diff
@@ -343,6 +343,57 @@ class ContributorChildAdmin(
...
+            return '-'
+        url = reverse('admin:activity_pub_organization_change', args=(platform.pk,))
+        return format_html('<a href="{}">{}</a>', url, platform.name)
+
+    platform.short_description = _('Platform')
+
+    def with_remote_user_field(self, fields, obj):
+        fields = list(fields)
+        has_remote = bool(obj and getattr(obj, 'remote_user_id', None))
+        if has_remote or self.activity_is_shared(obj):
```
WolfHakase, line 351: Showing remote_user whenever the activity is shared (since it's 'or') means empty remote-user fields on Organizer / Donor / Collect / etc can change pages. Is this correct behaviour or should this only happen when there is a remote (by changing to 'and')?

bluebottle/time_based/admin.py
```diff
@@ -331,6 +332,54 @@ class TeamMemberAdmin(RegionManagerAdminMixin, StateMachineAdmin):
...
+        return fields
+
+    def get_readonly_fields(self, request, obj=None):
+        fields = list(super().get_readonly_fields(request, obj))
+        has_remote = bool(obj and getattr(obj, 'remote_user_id', None))
+        if (has_remote or self.activity_is_shared(obj)) and 'remote_user' not in fields:
+            fields.append('remote_user')
+        if has_remote and 'platform' not in fields:
+            fields.append('platform')
+        return fields
```
WolfHakase: This block is nearly identital to ContributorChildAdmin.with_remote_user_field / platform / get_readonly_fields, the same also exists on TeamAdmin (~566–612) and RegistrationChildAdmin (~2014–2060). Is it worth extracting a mixin so these won't drift?

bluebottle/time_based/admin.py
```diff
@@ -343,31 +392,55 @@ def get_fieldsets(self, request, obj=None):
...
 
-    def link(self, obj):
+    def edit(self, obj):
+        if not obj.pk:
+            return '-'
+        if not obj.user_id and obj.team.activity.has_deleted_data:
+            return format_html('<i>{}</i>', _('Anonymous'))
         url = reverse('admin:time_based_teammember_change', args=(obj.id,))
-        return format_html('<a href="{}">{}</a>', url, obj)
+        return format_html('<a href="{}">{}</a>', url, _('Edit'))
```
WolfHakase: participant / edit mostly duplicate BaseParticipantUserInline, only switching obj.activity to obj.team.activity. Can we share that via a helper?

WolfHakase (changes requested):
If the behaviour is correct let me know, then I'll change it to approved. The suggested refactors are optional in my opinion.

bluebottle/activities/admin.py
gannetson: Yes this is correct. 
`if obj and obj.is_adopted:` id the correct condition to show a read_only `display_image` instead of full control `image`

PR #7127, opened 2026-08-13
Ticket/bb 29792 spot open email

WolfHakase:
Based on branch: ticket/bb-29787-notify-me-button-on-full-activity-slot https://github.com/onepercentclub/bluebottle/pull/7122. Recommend reviewing and merging first.

PR #7128, opened 2026-08-13
Ticket/bb 29793 spot opened email flow

WolfHakase:
Branch based on #7127. Recommend reviewing and merging first.


bluebottle/notifications/management/commands/preview_all_messages.py
```diff
@@ -259,14 +259,26 @@ def __init__(self, elements=None):
...
 
     def last(self):
-        return self.elements[-1]
+        return self.elements[-1] if self.elements else None
 
     def count(self):
         return len(self.elements)
 
+    def all(self):
+        return self
```
gannetson: Didn't read the rest of the code yet, but looking at the other methods here I would expect this tot return `self.items` and the next two to actualy apply `filter` and do `select_related`.

bluebottle/time_based/templates/mails/messages/spot_opened.html
```diff
@@ -0,0 +1,36 @@
...
+
+{% block message %}
+    <p>
+        {% blocktrans context 'platform-email' %}Good news, a spot has opened up for the activity:{% endblocktrans %}
+    </p>
+    <p>
+        <b>{{ title }}</b><br/>
+        {% if start or end %}
+            {% if start and end %}
+                {{ start }} – {{ end }}<br/>
```
gannetson: Not sure if these dates will format nicely, please check and otherwise remove it as we decided to leave this out in refinement.

bluebottle/time_based/triggers/slots.py
```diff
@@ -39,6 +42,14 @@
 from bluebottle.time_based.states.participants import DateParticipantStateMachine
 
 
+def slot_registration_deadline_is_not_passed(effect):
```
gannetson: This returns activity registration deadline, not 'slot_registration_deadline', please rename, to avoid confusion.

bluebottle/time_based/triggers/triggers.py
```diff
@@ -368,7 +385,9 @@ def activity_will_not_be_full(effect):
 
     return (
         not activity.capacity or
-        activity.capacity >= len(activity.accepted_participants)
+        activity.capacity > spots_taken_after_release(
+            activity.accepted_participants, effect.instance
+        )
```
gannetson, line 377: Can't we simplify this like we do at `activity_will_be_full`? We will change the finite state sometime soonish, where we will calculate states after all transitions have been made, then we don't have to do the weird '+1/-1' stuff.

bluebottle/time_based/messages/messages.py
```diff
@@ -938,3 +940,55 @@ def action_link(self):
...
+            url = self.obj.get_absolute_url()
+        return '{}?{}'.format(url, urlencode({'spotOpened': 'true'}))
+
+    def get_recipients(self):
+        """interested members"""
+        if isinstance(self.obj, DateActivitySlot):
+            return [interest.user for interest in self.obj.interests.all()]
+        return [
+            interest.user
+            for interest in self.obj.interests.filter(slot__isnull=True)
```
eodolphi: Shouldn't this filter on the specific slot:

`self.obj.interests.filter(slot=self.obj)`

bluebottle/time_based/serializers/interests.py
```diff
@@ -0,0 +1,158 @@
...
+            'user',
+            'activity',
+        ]
+
+    included_serializers = {
+        'user': 'bluebottle.initiatives.serializers.MemberSerializer',
+        'activity': 'bluebottle.activities.serializers.ActivitySerializer',
+    }
+
+    def _user_already_involved(self, user, activity, slot):
```
eodolphi: Small cleanup: Move this validation logic to a separate class, and possibly split it into multiple validators.

bluebottle/time_based/messages/messages.py
WolfHakase: In the refinement we decided that we would notify the user about the activity itself rather than the slot. If you want we can discuss with Rolf (on Monday) whether on slot makes more sense.

WolfHakase: Nvm the above. We discussed and now only the users who've shown interest in the specific slot get emailed.

eodolphi (approved):

bluebottle/notifications/management/commands/preview_all_messages.py
WolfHakase: fixed

bluebottle/time_based/triggers/slots.py
WolfHakase: fixed

bluebottle/time_based/templates/mails/messages/spot_opened.html
WolfHakase: fixed

bluebottle/time_based/triggers/triggers.py
WolfHakase: I agree that once the FSM is fixed this should be replaced. But, the simplification of activity_will_be_full is not equalilty imprecise. It would bvreak a deliberate and tested product rule. See SpotsTakenAfterReleaseTestCase. If you still think simplifying is better, I'm happy to comply.

PR #7130, opened 2026-08-14
Only reviewers should be allowed to cancel a grant application

eodolphi:


WolfHakase (approved):
Would be nice if we could add a regression test. But looks good to me.

PR #7131, opened 2026-08-14
Ticket/bb 29794 manager interested list

WolfHakase:


bluebottle/time_based/serializers/interest_link_field.py
```diff
@@ -0,0 +1,80 @@
...
+
+def _get_owners(instance):
+    owners = getattr(instance, 'owners', None)
+    if owners is None and hasattr(instance, 'activity'):
+        owners = getattr(instance.activity, 'owners', None)
+    return owners
+
+
+def remove_interests_field_for_non_managers(serializer, instance):
+    if not isinstance(instance, Model) or 'interests' not in serializer.fields:
```
eodolphi: Maybe explain why this is needed here.

bluebottle/time_based/serializers/interest_link_field.py
```diff
@@ -0,0 +1,80 @@
...
+            return False
+
+        return (
+            request.user in owners or
+            request.user.is_staff or
+            request.user.is_superuser
+        )
+
+    def get_attribute(self, instance):
+        if not self._can_view_interests(instance):
```
eodolphi: Isn't this enough for not sending interests to non-managers? And even then, isn't the important part that you cannot read the link if you are a non-manager? Sending the link to normal users does not reveal any information. They can figure out the link without the help of the backend even.

eodolphi (changes requested):

bluebottle/time_based/serializers/interest_link_field.py
WolfHakase: You're right. This isn't needed. Resolved.

bluebottle/time_based/serializers/interest_link_field.py
WolfHakase: Correct again. Resolved.

eodolphi (approved):

PR #7132, opened 2026-08-18
Ticket/bb 29795 export interested participants

WolfHakase:


bluebottle/time_based/views/exports.py
```diff
@@ -16,6 +16,14 @@
 from bluebottle.utils.admin import prep_field
 from bluebottle.utils.views import ExportView
 
+INTEREST_EXPORT_FIELDS = (
+    ('user__email', 'Email'),
+    ('user__full_name', 'Name'),
+    ('created', 'Registration Date'),
+    ('status', 'Status'),
+)
+INTEREST_STATUS = 'Interested'
```
WolfHakase: The ticket explicitly asked for this status. But in reality the Interest model doesn't have a status. So it's hardcoded. I don't see the value, but it's an acceptance criteria...

eodolphi: If we do not see the value, we should probably just get rid of it.

WolfHakase: Agreed!

eodolphi (approved):

PR #7134, opened 2026-08-18
Cleanup serializers

gannetson:


bluebottle/activity_pub/serializers/__init__.py
```diff
@@ -86,12 +86,24 @@ def __init__(self, *args, **kwargs):
...
         self.resource_type_model_mapping['Team'] = LocalTeam
         self.resource_type_model_mapping['TeamScheduleSlot'] = TeamScheduleSlot
 
     def _get_resource_type_from_mapping(self, mapping):
         resource_type = super()._get_resource_type_from_mapping(mapping)
+        if resource_type == 'Join':
+            from bluebottle.activity_pub.models import ActivityPubModel, Team as ActivityPubTeam
+            from bluebottle.activity_pub.utils import resource_iri
+
+            object_ref = mapping.get('object')
```
eodolphi, line 99: This looks weird to me. Why does this have multiple ways of returning "TeamMember".

bluebottle/activity_pub/serializers/base.py
```diff
@@ -218,8 +219,24 @@ def to_internal_value(self, data):
 
         return super().to_internal_value(data)
 
+    def get_queryset(self):
+        return self.Meta.model.objects.all()
```
eodolphi: Isn't this the default?

bluebottle/activity_pub/serializers/base.py
```diff
@@ -218,8 +219,24 @@ def to_internal_value(self, data):
...
+
+    def save(self, **kwargs):
+        if self.instance is None:
```
eodolphi, line 223: This also look like the default. And we cannot really rely on the url_kwarg in AP. But since everything has a url and this url is stored, we can easily find the correct object to update.

bluebottle/activity_pub/serializers/base.py
```diff
@@ -238,7 +255,22 @@ def create(self, validated_data):
 
                         validated_data[field.source] = field.create(field_data)
 
-        result = super().create(validated_data)
+        model_class = self.get_queryset().model
```
eodolphi: This also looks a lot like the default, can't this use super?

bluebottle/activity_pub/serializers/fields.py
```diff
@@ -50,3 +51,10 @@ def to_representation(self, value):
 
     def to_internal_value(self, value):
         return {'type': self.type}
+
```
eodolphi, line 54: This looks like it is not used.

bluebottle/activity_pub/serializers/federated_activities.py
```diff
@@ -9,21 +9,24 @@
 from django.core.files import File
```
eodolphi, line 9: This file looks a lot better.

bluebottle/activity_pub/tests/test_team_add_authorization.py
```diff
@@ -6,15 +6,15 @@
...
+    TeamMemberJoinSerializer, TeamScheduleSlotsSerializer,
 )
 from bluebottle.activity_pub.tests.factories import OrganizationFactory, PersonFactory
 from bluebottle.cms.models import SitePlatformSettings
 from bluebottle.test.utils import BluebottleTestCase
 from bluebottle.time_based.tests.factories import ScheduleActivityFactory, TeamFactory
 
 
-class TeamMemberAddAuthorizationTestCase(BluebottleTestCase):
+class TeamMemberJoinAuthorizationTestCase(BluebottleTestCase):
```
eodolphi: I think I commented this before, but this does not check any authorization. These tests should probably live somewhere else.

bluebottle/activity_pub/tests/test_team_registration_identity.py
```diff
@@ -115,7 +120,91 @@ def test_registration_without_team_raises_validation_error(self):
...
             self.serializer.create({
-                'activity': self.slot,
+                'activity': self.activity,
+                'slot': self.slot,
                 'remote_user': self.remote_user,
             })
         self.assertIn('no team', str(error.exception.detail).lower())
+
+
+class TeamScheduleRegistrationSerializerCreateTestCase(BluebottleTestCase):
```
eodolphi, line 130: I am not sure how this test can actual test this, since it only seems to use one tenant.

bluebottle/activity_pub/tests/test_team_registration_identity.py
```diff
@@ -115,7 +120,91 @@ def test_registration_without_team_raises_validation_error(self):
...
+        )
+        registration = self.serializer.create({
+            'activity': self.activity,
+            'remote_user': self.remote_user,
+            'user': None,
+            'team': ap_team.iri,
+        })
+        ap_team.refresh_from_db()
+        self.assertEqual(ap_team.adopted_id, registration.teams.get().id)
+        self.assertEqual(ap_team.attributed_to_id, self.activity.activity_pub_model.id)
```
eodolphi: I thought we fixed this already, but attributed_to is not the right field to use for a team.

bluebottle/activity_pub/models.py
```diff
@@ -645,6 +645,15 @@ class Team(ActivityPubModel):
         related_name='origin',
     )
 
+    def save(self, *args, **kwargs):
+        if not self.attributed_to_id:
```
eodolphi: This should not be nessecary. The team should already be set, and we should not use attributed_to

bluebottle/activity_pub/models.py
```diff
@@ -1037,55 +1046,35 @@ def save(self, *args, **kwargs):
         super().save(*args, **kwargs)
 
         if not self.is_local:
-            adapter.adopt(self)
+            adopted = adapter.adopt(self)
```
eodolphi: I do not like doing model specific things here. This means that calling `adapter.adopt` from outside this model will not work correctly. Instead the adoption process should make sure the status is correct.

bluebottle/activity_pub/utils.py
```diff
@@ -73,3 +73,46 @@ def timedelta_to_iso(td):
...
+        return value
+    if isinstance(value, dict):
+        return value.get('id') or value.get('iri')
+    return None
+
+
+def event_for_team(team):
+    from bluebottle.activity_pub.models import Team as ActivityPubTeam
+
+    ap_team = getattr(team, 'origin', None)
```
eodolphi: This should not be nessecary when teams work like other participants.

bluebottle/activity_pub/serializers/federated_activities.py
```diff
@@ -1285,310 +1196,185 @@ def create(self, validated_data):
...
-
-        target = self.initial_data.get('target')
-        if isinstance(target, str):
-            target = {'id': target}
-        elif isinstance(target, dict) and 'id' not in target and 'iri' in target:
-            target = {'id': target['iri']}
-
-        target_iri = self._resource_iri(target)
-        if not target_iri:
+        if not platform_may_modify_event(platform, event):
```
eodolphi: This check should happen in the view layer, not this much down the serializer tree.

bluebottle/activity_pub/serializers/federated_activities.py
```diff
@@ -1285,310 +1196,185 @@ def create(self, validated_data):
...
 
         event_activity = event.origin if event.is_local else event.adopted
         if event_activity is None or event_activity.pk != team.activity_id:
             raise serializers.ValidationError({
-                'target': 'Add target team does not belong to the attributed activity',
+                'object': 'Join object team does not belong to the attributed activity',
             })
 
-        platform = self._get_add_platform()
+        request = self.context.get('request') if self.context else None
```
eodolphi: Why is this serializer the only one that needs to do this check?

bluebottle/time_based/effects/participants.py
```diff
@@ -149,21 +149,41 @@ class CreateDateRegistrationEffect(Effect):
...
-        if not self.instance.registration and self.instance.user:
-            self.instance.registration = DateRegistration.objects.create(
-                activity=self.instance.activity,
-                user=self.instance.user,
-            )
-
-            self.instance.save()
+        if self.instance.registration:
+            return
+        if not self._registration_filters():
```
eodolphi: This is a bit confusing. Just check if there is a user or not, since this hides the intention.

eodolphi (changes requested):

PR #7139, opened 2026-08-19
Ticket/bb 29799 overview activities interested

WolfHakase:


bluebottle/time_based/tests/test_interest_api.py
```diff
@@ -29,6 +32,37 @@
 )
 
 
+class InterestSerializerIncludesTestCase(SimpleTestCase):
+    """
+    Guard against regressions where compound includes are dropped from the
+    shared InterestSerializer when adding new interest list endpoints.
+    """
```
eodolphi: This is probably already tested in the api test below right? In that case I prefer the proper API test that actually verifies that this works.

eodolphi (approved):

PR #7145, opened 2026-08-27
Fix matching email preview

gannetson:


bluebottle/notifications/management/commands/preview_all_messages.py
WolfHakase, line 1: If we keep adding to this command, we may want to clean up the code a bit and separate out some of the helpers and json out of this file. It's growing a bit out of proportions. But out of scope.

WolfHakase (approved):

PR #7151, opened 2026-08-28
link the spot-opened email to the time slot that opened, and include …

WolfHakase:
…that slot’s date in the message

### DIFF
--- a/bluebottle/time_based/messages/messages.py
+++ b/bluebottle/time_based/messages/messages.py
@@ -956,34 +956,40 @@ class SpotOpenedNotification(TransitionMessage):
 
     @property
     def action_link(self):
-        if isinstance(self.obj, DateActivitySlot):
-            url = self.obj.activity.get_absolute_url()
-        else:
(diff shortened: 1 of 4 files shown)


bluebottle/time_based/messages/messages.py
```diff
@@ -956,34 +956,40 @@ class SpotOpenedNotification(TransitionMessage):
...
     @property
     def action_link(self):
-        if isinstance(self.obj, DateActivitySlot):
-            url = self.obj.activity.get_absolute_url()
-        else:
-            url = self.obj.get_absolute_url()
-        return '{}?{}'.format(url, urlencode({'spotOpened': 'true'}))
+        url = self.obj.get_absolute_url()
+        separator = '&' if '?' in url else '?'
+        return '{}{}{}'.format(url, separator, urlencode({'spotOpened': 'true'}))
```
eodolphi, line 961: I would rather parse the original url, but for now this works too.

eodolphi (approved):

PR #7152, opened 2026-09-01
Fix admin platform name in second place too

eodolphi:


gannetson (approved):

bluebottle/common/templates/admin/base_site.html
```diff
@@ -4,5 +4,9 @@
 {% block title %}{{ title }} | {% trans 'Django site admin' %}{% endblock %}
 
 {% block branding %}
-    <h1 id="site-name">{{ TENANT.tenant.client_name|title }}!!!!!!!</h1>
+    <h1 id="site-name">
+        {% if platform_name %}
+            <span class="segment-label segment-label-{{ tenant.slug }}">{{platform_name}}</span>
```
gannetson: Wait? Why a class name based on slug?

eodolphi: Because we have that in the other place too. It probably does not do much any more.

eodolphi: But looking at this in more detail, it never showeed anything, and it looks ugly. I'll just remove the platform name here. I just removed it completely.

PR #7153, opened 2026-09-02
Feature | Voting

gannetson:


bluebottle/bluebottle_dashboard/templates/django_quill/widget.html
```diff
@@ -8,56 +8,135 @@
...
-
-            async function upload(file) {
-                const response = await fetch(
-                    '/api/files/images',
-                    {
-                        method: "POST",
-                        body: file,
-                        headers: {
-                            "Content-Disposition": 'attachement; filename="' + file.name + '"'
+            window.djq = window.djq || {};
```
eodolphi: Why are these changes nessecary now?  Can we add some comment on what does what? This is a lot of code that will break with upgrades and is very hard to read.

bluebottle/cms/templates/admin/cms/preview/default.html
```diff
@@ -0,0 +1 @@
+<p>{{ instance }}</p>
```
eodolphi: So every block preview will now show the instance? Why?

bluebottle/voting/admin.py
```diff
@@ -0,0 +1,140 @@
...
+
+class PollVoteInline(admin.TabularInline):
+    model = PollVote
+    extra = 0
+    raw_id_fields = ('owner',)
+    readonly_fields = ('created', 'updated')
+    fields = ('owner', 'option', 'created', 'updated')
+    ordering = ('-created',)
+
+    def formfield_for_foreignkey(self, db_field, request=None, **kwargs):
```
eodolphi: This looks like a weird way of filtering the queryset. We should not pass semi globals using the request. That will fail in weird ways.

bluebottle/voting/admin.py
```diff
@@ -0,0 +1,140 @@
...
+            (_('Details'), {'fields': self.get_fields(request, obj)}),
+        )
+        if request.user.is_superuser:
+            fieldsets += (
+                (_('Super admin'), {'fields': self.superadmin_fields}),
+            )
+        return fieldsets
+
+    def get_formsets_with_inlines(self, request, obj=None):
+        request._poll_obj = obj
```
eodolphi: We should not pass things using the request

bluebottle/voting/admin.py
```diff
@@ -0,0 +1,140 @@
...
+        return obj.votes.count()
+
+    vote_count.short_description = _('votes')
+
+    def pages(self, obj):
+        if not obj or not obj.pk:
+            return '-'
+
+        links = []
+        seen = set()
```
eodolphi: What does this protect against? Having a page with the same poll repeated?

eodolphi (approved):

bluebottle/bluebottle_dashboard/templates/django_quill/widget.html
gannetson: Tossed some things around so it is more structured and clearer. NB this also fixed a long existing anoying bug where in CMS pages you first have to save before you can enter something into Quill inputs

bluebottle/cms/templates/admin/cms/preview/default.html
gannetson: Will remove this

bluebottle/voting/admin.py
gannetson: This filter is here so that you cannot select options form other polls. I think we the clean-up of that weird hijacking of request this method is now ok.

bluebottle/voting/admin.py
gannetson: The hypothetical case where the same poll is placed multiple times on the same page :facepalm: 
Removed code

eodolphi (approved):

PR #7162, opened 2026-09-07
when an activity goes from closed to open email notification now gets…

WolfHakase:
… sent

### DIFF
--- a/bluebottle/time_based/effects/slots.py
+++ b/bluebottle/time_based/effects/slots.py
@@ -84,10 +84,7 @@ class ReopenRegistrationClosedSlotsEffect(Effect):
             return
 
         for slot in self.reopened_slots:
-            slot.save(run_triggers=False)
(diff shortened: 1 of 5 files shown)


bluebottle/time_based/effects/slots.py
```diff
@@ -85,6 +85,9 @@ def post_save(self, **kwargs):
 
         for slot in self.reopened_slots:
             slot.save(run_triggers=False)
+            while slot._postponed_effects:
```
eodolphi: Is there a reason why this needs to happen here. The base model has code to execute postponed effect automatically. Or might it be that we should run triggers on this save?

eodolphi (changes requested):

bluebottle/time_based/effects/slots.py
WolfHakase: Good point. Fixed.

PR #7163, opened 2026-09-07
Use hijack for impersonating

gannetson:


bluebottle/auth/middleware.py
```diff
@@ -349,6 +376,11 @@ def _verify_user(self, request, user):
                 else None
             )
 
+            if getattr(user, 'is_hijacked', False):
```
eodolphi: Except for a test I do not see this being set anywhere.

bluebottle/members/hijack.py
```diff
@@ -0,0 +1,12 @@
+def can_hijack(*, hijacker, hijacked):
```
eodolphi, line 1: Can this be in a more general and logical location?

eodolphi: When will this be called with None as one of the arguments?

bluebottle/members/utils.py
```diff
@@ -35,3 +35,29 @@ def jwt_payload_handler(user):
         payload['iss'] = api_settings.JWT_ISSUER
 
     return payload
+
+
+def jwt_is_impersonated(request):
```
eodolphi, line 40: Can we make this a middleware or part of another middleware? It is probably usefull to cache this on the request, because we want to access this in other places too.

eodolphi (changes requested):

PR #7175, opened 2026-09-14
Fix sending mail for some cases

gannetson:


bluebottle/clients/mail.py
```diff
@@ -18,10 +20,15 @@ def construct_from_header():
     # properties has not be initialised yet.
     settings = MailPlatformSettings.load()
```
eodolphi, line 19: What does this fix exactly?

eodolphi (approved):

PR #7176, opened 2026-09-14
Fix default owner on adopt

gannetson:


eodolphi:
Is this how we think it should work? If you adopt the activity, but there is a default owner, we use the default owner? I thought the default owner was meant for situations where there is no user in the request.

If we do this, we should rename the field to `acitivty_owner` or something. Or at least change the description of the default owner field to indicate that this is always the owner for newly adopted activities.

PR #7181, opened 2026-09-15
BB-30082: Make activity answers read-only on create

gannetson:
## Summary
- Duplicate-create was posting the source activity including its `answers` relationship. `BaseActivitySerializer` treated that reverse FK as writable, so `perform_create` blew up with `TypeError: Direct assignment to the reverse side of a related set is prohibited. Use answers.set() instead.` and, if it had succeeded, would have stolen the original rows.
- Mark `answers` `read_only=True` so create ignores them. The copy starts with zero answers and still reports required questions as missing; the wizard should POST them to `/activities/answers/` afterwards.
- Deed list API tests cover 201, source answers staying put, and the duplicate remaining incomplete.

## Test plan
- [x] `DeedsListViewAPITestCase` including the three new duplicate tests
- [x] `bluebottle.deeds` + `bluebottle.activities` (267 tests)
- [ ] Duplicate a deed (and another activity type) in the UI; create should not 500
- [ ] Original activity still has its answers; copy has none and required questions show as missing
- [ ] After posting answers on the copy, it can be published

Made with [Cursor](https://cursor.com)


eodolphi (changes requested):
This will work to get rid of the error. But.... Are we sure that the frontend does post  the questions / answers?

